PR #52, opened 2019-04-05
fix the wrong unit in CreateVolume

lizebang:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespaces from that line:

/kind bug

The unit used in CreateVolume should be GiB, not Byte.

Fixes #

**Release note**:
```
fix the wrong unit in CreateVolume
```


ashishranjan738:
good find @lizebang . LGTM

lizebang:
Thanks @ashishranjan738

PR #64, opened 2019-05-11
fix secret value exposed in logs issue

### LINKED ISSUE #60: secret value should not be exposed in logs
**What happened**:
Currently secret value would be exposed in logs, while it should not

```
I0421 02:40:19.501807       1 utils.go:112] GRPC response: capabilities:<rpc:<> >
I0421 02:40:19.506657       1 utils.go:106] GRPC call: /csi.v1.Node/NodePublishVolume
I0421 02:40:19.506674       1 utils.go:107] GRPC request: volume_id:"arbitrary-volumeid" target_path:"/var/lib/kubelet/pods/cc9552fd-63de-11e9-8b22-000d3a004ffb/volumes/kubernetes.io~csi/pv-blobfuse/mount" volume_capability:<mount:<> access_mode:<mode:MULTI_NODE_MULTI_WRITER > > secrets:<key:"accountkey" value:"xxx" > secrets:<key:"accountname" value:"andyacidiag" > volume_context:<key:"containerName" value:"test" > volume_context:<key:"resourceGroup" value:"andy-aci" > volume_context:<key:"storageAccount" value:"andyacidiag" >
```

We could disable this logging easily, while I would like to filter out only the `secret` value, for the other request info, it's useful for debugging:
https://github.com/kubernetes-sigs/azurefile-csi-driver/blob/224a7d36921cc7ce354cef12c9e8ae791cc1371a/pkg/csi-common/utils.go#L107

**How to reproduce it**:

**Environment**:
- CSI Driver version: v0.2.0-alpha
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

11janci:
**What type of PR is this?**
/kind bug

Mask secrets in logs.

Fixes #60 

**Release note**:
```
fix secret value exposed in logs issue
```

### DIFF
--- a/pkg/csi-common/utils.go
+++ b/pkg/csi-common/utils.go
@@ -18,12 +18,12 @@ package csicommon
 
 import (
 	"fmt"
-	"strings"
-
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	"golang.org/x/net/context"
 	"google.golang.org/grpc"
 	"k8s.io/klog"
+	"regexp"
+	"strings"
 )
 
 func ParseEndpoint(ep string) (string, string, error) {
@@ -102,9 +102,15 @@ func RunControllerandNodePublishServer(endpoint string, d *CSIDriver, cs csi.Con
 	s.Wait()
 }
 
+// regex to mask secrets in log messages
+var reqSecretsRegex, _ = regexp.Compile("secrets\\s*:\\s*<key:\"(.*?)\"\\s*value:\".*?\"")
+
 func logGRPC(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+
+	s := fmt.Sprintf("GRPC request: %+v", req)
 	klog.V(3).Infof("GRPC call: %s", info.FullMethod)
-	klog.V(5).Infof("GRPC request: %+v", req)
+	klog.V(5).Info(reqSecretsRegex.ReplaceAllString(s, "secrets:<key:\"$1\" value:\"****\""))
+
 	resp, err := handler(ctx, req)
 	if err != nil {
 		klog.Errorf("GRPC error: %v", err)
--- a/pkg/csi-common/utils_test.go
+++ b/pkg/csi-common/utils_test.go
@@ -17,8 +17,14 @@ limitations under the License.
 package csicommon
 
 import (
+	"bytes"
+	"context"
+	"flag"
+	"google.golang.org/grpc"
+	"k8s.io/klog"
 	"testing"
 
+	"github.com/container-storage-interface/spec/lib/go/csi"
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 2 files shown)


pkg/csi-common/utils.go
```diff
@@ -102,9 +102,22 @@ func RunControllerandNodePublishServer(endpoint string, d *CSIDriver, cs csi.Con
...
+
 func logGRPC(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+
+	containsSecrets, origSecrets := maskSecrets(req)
+
 	klog.V(3).Infof("GRPC call: %s", info.FullMethod)
 	klog.V(5).Infof("GRPC request: %+v", req)
+
+	if containsSecrets {
+		unmaskSecrets(req.(requestWithSecrets), origSecrets)
```
andyzhangx: is it possible to copy the `req` variable, so we don't need to do `unmaskSecrets`, original way will change the `req` value, there is risk for doing this.
Another option is change print way for below statement, that would be safer
```
klog.V(5).Infof("GRPC request: %+v", req)
```

11janci: the problem with the copy is that the Secrets field will still point to the same map, so wouldn't make any difference - would have to do a deep copy, which I prefer to avoid if possible.
Other options I thought about are string replace, which is quite fragile IMO; and reflection, which is also generally discouraged. So while I agree that mutating the original object is not ideal, it occurred to me as the best viable option. But if you see another one, or you'd prefer either the string replacement or the reflection let me know and I will change it.
thx for the review :)

11janci:
Hi @andyzhangx,
will take a look and fix those issues in the test tomorrow.
Thx

pkg/csi-common/utils.go
andyzhangx: maybe just write a print function to replace `klog.V(5).Infof("GRPC request: %+v", req)`, change the original value is not a good idea, it may have potential data corruption issue.

11janci: do you mean using a string replace function/regex? for example something like
```
s := fmt.Sprintf("%+v", req)
... replace secrets in s using regex ...
```
I don't think anything else (outside of a copy + reflection) is possible.
copy + reflection:
```
reqCopy := copy // makes a shallow copy of the req
... use reflection to replace the secrets map in reqCopy with a new map with masked secrets ...
```
From those two I like the reflection approach more , is more robust. String replacement/regex is not very reliable imo.

andyzhangx: `String replacement/regex` and reflection approach are both ok for me, don't change the `req` value is the bottom line IMO. Thanks.

11janci:
@andyzhangx,
pls take a look at the updated PR. I ended up going the regex way, is much simpler code and faster (about 3 seconds difference for 500 000 requests compared to reflection).
Let me know what you think.
Thx

PR #121, opened 2019-09-28
feat: support VolumeExpansion in OFFLINE scenario

ZeroMagic:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
support VolumeExpansion in OFFLINE scenario

**Which issue(s) this PR fixes**:

Fixes #43 

**Special notes for your reviewer**:

**Release note**:
```
feat: support VolumeExpansion in OFFLINE scenario
```

Issue #18, opened 2019-02-15
[feature] Support snapshot & restore 


andyzhangx:
https://kubernetes-csi.github.io/docs/snapshot-restore-feature.html#snapshot--restore-feature

andyzhangx:
needs to figure out where is the go sdk of azure file snapshot first: https://github.com/Azure/azure-sdk-for-go/issues/4093

andyzhangx:
here is the azure file snapshot example:
https://github.com/Azure/azure-storage-file-go/blob/d93e19c84c2af7958331fb3ec8fd04eb0d76a25e/azfile/zt_examples_test.go#L521-L602

tall49er:
Can I working on it?

andyzhangx:
@tall49er sure, thanks.

ZeroMagic:
Wait for Storage Team to release the new feature, clone snapshot (create new file share from snapshot).

PR #122, opened 2019-10-07
chore: Update csi-sanity

ZeroMagic:
**What type of PR is this?**
/kind bug
/kind flake

In order to test `EXPAND_VOLUME` controller capability, we need to update the version of CSI sanity test to v2.2.0. 
Ref: #121 

Here I fix some bugs to pass the sanity tests.
- According to the logic of sanity test `should work` in `Node Service`, we should delete the target directory in the end of `func NodeUnpublishVolume`.

- Since we enabled `csi.ControllerServiceCapability_RPC_CREATE_DELETE_SNAPSHOT`, we should check the snapshot if it exists when creating the volume. Currently, AzureFile does not support restore. Ref: https://github.com/kubernetes-sigs/azurefile-csi-driver/issues/18#issuecomment-529316027 . So I do not make the code of `creating volume from snapshot` stand-alone as a new function, just include it in `CreateVolume`. 

Fixes #

**Release note**:
```
Update csi-sanity
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -168,7 +168,7 @@ replace (
 	github.com/kr/pretty => github.com/kr/pretty v0.1.0
 	github.com/kr/pty => github.com/kr/pty v1.1.5
 	github.com/kr/text => github.com/kr/text v0.1.0
-	github.com/kubernetes-csi/csi-test => github.com/kubernetes-csi/csi-test v1.1.0
+	github.com/kubernetes-csi/csi-test => github.com/kubernetes-csi/csi-test v2.2.0+incompatible
 	github.com/kubernetes-csi/external-snapshotter => github.com/kubernetes-csi/external-snapshotter v1.0.1
 	github.com/mailru/easyjson => github.com/mailru/easyjson v0.0.0-20190312143242-1de009706dbe
 	github.com/marstr/guid => github.com/marstr/guid v1.1.0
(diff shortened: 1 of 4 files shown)


pkg/azurefile/controllerserver.go
```diff
@@ -92,19 +92,41 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
...
+			return nil, status.Errorf(codes.Internal, "failed to check if snapshot(%v) exists: %v", snapshotId, err)
+		}
+		if !exists {
+			return nil, status.Errorf(codes.NotFound, "does not find the snapshot(%v)", snapshotId)
+		}
+
+		// Currently, azurefile doesn't support create volume from snapshot.
+		// https://github.com/kubernetes-sigs/azurefile-csi-driver/issues/18
+		// So the driver will create an empty fileshare for the user.
+		klog.Errorf("doesn't support create volume from snapshot, so create an empty file share")
```
andyzhangx: if we don't support `create volume from snapshot`, why not return error?

andyzhangx:
also, this PR is to fix the expand volume issue, why change the logic of snapshot functionality in this PR?

pkg/azurefile/controllerserver.go
ZeroMagic: Because we should follow the new sanity tests. Otherwise, the new sanity tests will fail.

- should create volume from an existing source snapshot

- should fail when the volume source snapshot is not found

ZeroMagic:
> also, this PR is to fix the expand volume issue, why change the logic of snapshot functionality in this PR?

This PR is to update sanity tests and make them work well.

pkg/azurefile/controllerserver.go
andyzhangx: while we don't support that, can we disable this sanity test case?

ZeroMagic: yes. we can skip these two tests.

PR #157, opened 2020-01-23
chore: set replica of controller as 2

andyzhangx:
**What type of PR is this?**
/kind cleanup

chore: set replica of controller as 2

Fixes #

**Release note**:
```
none
```


charts/latest/azurefile-csi-driver/templates/csi-azurefile-controller.yaml
```diff
@@ -22,6 +22,7 @@ spec:
         - name: csi-provisioner
           image: "{{ .Values.image.csiProvisioner.repository }}:{{ .Values.image.csiProvisioner.tag }}"
           args:
+            - "-v=5"
```
feiskyer: 5 seems too verbose, how about using 3?

andyzhangx: I would suggest as 5 to provide more detailed logs, `csi-provisioner` only do the provision work, it won't have too many logs.

PR #189, opened 2020-03-08
chore: change volumeid format to include diskname

andyzhangx:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
chore: change volumeid format to include diskname

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Release note**:
```
none
```

PR #214, opened 2020-03-20
fix: volumeid compatibility issue

andyzhangx:
**What type of PR is this?**
/kind bug

PR(https://github.com/kubernetes-sigs/azurefile-csi-driver/pull/189) introduced a volumeid compatibility issue, this PR is to fix that issue.

Fixes #

cc @aelmanaa

**Release note**:
```
fix: volumeid compatibility issue
```


andyzhangx:
the timeout would cause panic:
https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_azurefile-csi-driver/214/pull-azurefile-csi-driver-e2e-vmss/1240871465306296322/build-log.txt
Could you take a look? @ZeroMagic 
if it's timeout, then that case should fail instead of panic

ZeroMagic:
It seems that the problem is at `kubernetes/test/e2e/framework/util.go`, not related to our code. We can decrease the priority of it. I will deeply investigate it later.

PR #231, opened 2020-04-03
chore: remove cluster driver registrar

ZeroMagic:
**What type of PR is this?**
/kind cleanup

According to Kubernetes CSI Developer Documentation, the cluster-driver-registrar side-car which was used to create CSIDriver objects in Kubernetes 1.13 has been deprecated for Kubernetes 1.16. No cluster-driver-registrar has been released for Kubernetes 1.14 and later.

Fixes #

**Release note**:
```
remove cluster driver registrar
```


andyzhangx:
could you make sure this PR works in k8s 1.14? There are quite a lot users are using 1.14 now, we need to make sure it works. thanks  @ZeroMagic

deploy/csi-azurefile-driver.yaml
```diff
@@ -0,0 +1,10 @@
+---
+apiVersion: storage.k8s.io/v1beta1
+kind: CSIDriver
+metadata:
+  name: file.csi.azure.com
+spec:
+  attachRequired: true
+  podInfoOnMount: true
+  volumeLifecycleModes:  # added in Kubernetes 1.16
```
andyzhangx, line 9: pls remove this, there are users are still using 1.16, we don't want to break those users.

ZeroMagic: done

PR #248, opened 2020-04-23
test: fix windows e2e test failure

andyzhangx:
**What type of PR is this?**
/kind test

**What this PR does / why we need it**:
fix windows e2e test failure, this is just a workaround since `mcr.microsoft.com/dotnet/framework/samples:aspnetapp` is 1.9GB, too large

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Release note**:
```
none
```

PR #249, opened 2020-04-23
fix: remove unnecessary SharedInformerFactory in unit tests

ZeroMagic:
**What type of PR is this?**
/kind bug

remove unnecessary SharedInformerFactory in unit tests.

Fixes #

**Release note**:
```

```

### DIFF
--- a/pkg/azurefile/utils_test.go
+++ b/pkg/azurefile/utils_test.go
@@ -17,17 +17,13 @@ limitations under the License.
 package azurefile
 
 import (
-	"context"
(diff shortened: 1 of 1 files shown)


ZeroMagic:
Because the last merged PR(https://github.com/kubernetes-sigs/azurefile-csi-driver/pull/248) used the first unit tests result as the coverage rate. But the first result is incomplete due to the failure. So it is higher than the normal one.

Here we should ignore the decrease in coverage rate.

Issue #258, opened 2020-04-27
improve unit test coverage


andyzhangx:
**Is your feature request related to a problem?/Why is this needed**

**Describe the solution you'd like in detail**

improve UT coverage including
 - controllerserver
 - nodeserver
 - identityserver
 - version

https://coveralls.io/github/kubernetes-sigs/azurefile-csi-driver?branch=master

![image](https://user-images.githubusercontent.com/4178417/80330808-d1785a00-8878-11ea-9529-eb7eb03cfa10.png)

**Describe alternatives you've considered**

**Additional context**

Dhiraj240:
@andyzhangx Would like to work on it.

andyzhangx:
> @andyzhangx Would like to work on it.

thanks, this is a big part, just update which part your are working on this issue when you have any progress.

andyzhangx:
and there are some easy parts, e.g. `version.go`, this issue is to improve the test code coverage, current ratio is awful, thanks.

andyzhangx:
cc @kushthedude to take a look, thanks

animeshk08:
Working on unit tests for https://github.com/kubernetes-sigs/azurefile-csi-driver/blob/master/pkg/azurefile/nodeserver.go. Mentioning to avoid redundant work :)

animeshk08:
Update:

Uncovered files:
* `server.go`: Need to find a way to mock grpc requests.

Improve coverage:
 * `controllerserver.go`: Need to create mock file client.
* `nodeserver.go`: Mock mount/unmount operation to remove the need of running the command with root privileges.

andyzhangx:
thanks @animeshk08 really impressive work!  
Can you improve code coverage on `csi-common`? thanks

Current code coverage:
![image](https://user-images.githubusercontent.com/4178417/84569999-f68b4280-adbc-11ea-95b3-6ab5252027ca.png)

cc @Sakuralbj

animeshk08:
Sure!
`controllerserver-default.go` has unimplemented methods which I have omitted. Hence the low coverage for that file.
I will start work on `server.go`.

andyzhangx:
it's now 80%!
![image](https://user-images.githubusercontent.com/4178417/86123809-b93ce980-bb0c-11ea-8785-572be73f353a.png)

andyzhangx:
it's now 85%, thanks for the contribution! @animeshk08  now close this issue.
[...]

PR #265, opened 2020-05-06
Added test for version.go and resolved undefined error in that same file

Dhiraj240:
Need reviews on the tests.
Tests are on version file
#258 
#266

### DIFF
--- a/pkg/azurefile/version_test.go
+++ b/pkg/azurefile/version_test.go
@@ -1,0 +1,62 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package azurefile
+
+import (
+	"fmt"
+	"reflect"
+	"runtime"
+	"strings"
+	"testing"
+
+	"sigs.k8s.io/yaml"
+)
+
+func TestGetVersion(t *testing.T) {
+	version := GetVersion()
+
+	expected := VersionInfo{
+		DriverName:    "azurefile-csi-driver",
+		DriverVersion: "N/A",
+		GitCommit:     "N/A",
+		BuildDate:     "N/A",
+		GoVersion:     runtime.Version(),
+		Compiler:      runtime.Compiler,
+		Platform:      fmt.Sprintf("%s/%s", runtime.GOOS, runtime.GOARCH),
+	}
+
+	if !reflect.DeepEqual(version, expected) {
+		t.Fatalf("structs are not same\ngot:\n%+v\nexpected:\n%+v", version, expected)
+	}
+
+}
+
+func TestGetVersionYAML(t *testing.T) {
+	output, err := GetVersionYAML()
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+
+	versioninfo := GetVersion()
+	marshalled, _ := yaml.Marshal(&versioninfo)
+
+	expected := strings.TrimSpace(string(marshalled))
+
+	if output != expected {
+		t.Fatalf("yaml not same\ngot:\n%s\nexpected:\n%s", output, expected)
+	}
+}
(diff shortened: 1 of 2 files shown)


andyzhangx:
@Dhiraj240 this test failed:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_azurefile-csi-driver/265/pull-azurefile-csi-driver-unit/1258204236047454213

Dhiraj240:
Yes, it was due to changes I made to use structure in order to have efficiency but couldn't work. I will change it.

pkg/azurefile/version_test.go
```diff
@@ -0,0 +1,61 @@
...
+		t.Log("Got as expected")
+	} else {
+		t.Errorf("Unexpected Versions: Please check the requirements")
+	}
+	
+}
+
+func TestGetVersionYAML(t *testing.T) {
+	get := GetVersion()
+	response, err := yaml.Marshal(&get)
```
animeshk08, line 54: Hi @Dhiraj240. I think your test is failing due to this line. Try changing response to _

animeshk08:
Also, consider running `hack/update-gofmt.sh` before pushing the changes. It formats the Go source code :)

Dhiraj240:
> Also, consider running `hack/update-gofmt.sh` before pushing the changes. It formats the Go source code :)

It says no  No such file or directory

animeshk08:
This is the file: https://github.com/kubernetes-sigs/azurefile-csi-driver/blob/master/hack/update-gofmt.sh

Dhiraj240:
> > It says no No such file or directory
> 
> This is the file: https://github.com/kubernetes-sigs/azurefile-csi-driver/blob/master/hack/update-gofmt.sh

I got this update-gofmt.sh: 17: set: Illegal option -o pipefail

Dhiraj240:
Moreover I checked it using below and got error
bash -x  update-gofmt.sh
+ set -euo pipefail
+ find . -name '*.go'
+ grep -v '\/vendor\/'
+ xargs gofmt -s -w
error: cannot use -w with standard input

Dhiraj240:
@animeshk08 Can you share your slack contact or any other?

pkg/azurefile/version_test.go
Dhiraj240: Oh, sorry.

Dhiraj240:
@andyzhangx Test slightly increased coverage but stands done as of now.

pkg/azurefile/version_test.go
```diff
@@ -0,0 +1,61 @@
...
+package azurefile
+
+import (
+	"testing"
+
+	"sigs.k8s.io/yaml"
+)
+
+func TestGetVersion(t *testing.T) {
+	get := VersionInfo{
```
andyzhangx, line 26: You are not testing `GetVersion` func

pkg/azurefile/version_test.go
```diff
@@ -0,0 +1,61 @@
...
+	if get == want {
+		t.Log("Got as expected")
+	} else {
+		t.Errorf("Unexpected Versions: Please check the requirements")
+	}
+
+}
+
+func TestGetVersionYAML(t *testing.T) {
+	get := GetVersion()
```
andyzhangx, line 53: Should test against `GetVersionYAML` func

Dhiraj240: Changes done

pkg/azurefile/version_test.go
Dhiraj240: Tested now, without structure approach

pkg/azurefile/version_test.go
```diff
@@ -0,0 +1,56 @@
...
+	if get == want {
+		t.Log("Got as expected")
+	} else {
+		t.Errorf("Unexpected Versions: Please check the requirements")
+	}
+
+}
+
+func TestGetVersionYAML(t *testing.T) {
+	_, get := GetVersionYAML()
```
andyzhangx, line 48: it's actually should be like this:
```
yaml, err := GetVersionYAML()
```
and you should compare the `yaml` with expected string and also check there is no `err`

pkg/azurefile/version_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"runtime"
+	"testing"
+
+	"sigs.k8s.io/yaml"
+)
+
+func TestGetVersion(t *testing.T) {
+	get := GetVersion()
+
+	want := VersionInfo{
```
andyzhangx, line 30: `want` --> `expected`

pkg/azurefile/version_test.go
Dhiraj240: alright

Issue #119, opened 2019-09-16
handle ShareBeingDeleted error when DeleteVolume


andyzhangx:
**What happened**:

We should handle below `ShareBeingDeleted` in DeleteVolume function, need to retry:

```
I0912 23:15:06.139324   12105 utils.go:111] GRPC call: /csi.v1.Controller/DeleteVolume
I0912 23:15:06.139355   12105 utils.go:112] GRPC request: volume_id:"#blobfuseci#sanity-node-full-8b4579a6-63d6317e" 
E0912 23:15:06.298687   12105 utils.go:116] GRPC error: rpc error: code = Internal desc = DeleteFileShare sanity-node-full-8b4579a6-63d6317e under blobfuseci failed with error: -> github.com/Azure/azure-storage-file-go/azfile.newStorageError, /home/travis/gopath/pkg/mod/github.com/!azure/azure-storage-file-go@v0.5.0/azfile/zc_storage_error.go:42
===== RESPONSE ERROR (ServiceCode=ShareBeingDeleted) =====
Description=The specified share is being deleted. Try operation later.
RequestId:58c3b74a-f01a-0185-5fbf-69ba2e000000
Time:2019-09-12T23:15:06.2888953Z, Details: 
   Code: ShareBeingDeleted
   DELETE https://blobfuseci.file.core.windows.net/sanity-node-full-8b4579a6-63d6317e?restype=share&timeout=61
   Authorization: REDACTED
   User-Agent: [Azure-Storage/0.5.0 (go1.12.1; linux)]
   X-Ms-Client-Request-Id: [e1117938-be6a-44eb-4ea5-290f123e9650]
   X-Ms-Date: [Thu, 12 Sep 2019 23:15:06 GMT]
   X-Ms-Delete-Snapshots: [include]
   X-Ms-Version: [2018-03-28]
   --------------------------------------------------------------------------------
   RESPONSE Status: 409 The specified share is being deleted. Try operation later.
   Content-Length: [244]
   Content-Type: [application/xml]
   Date: [Thu, 12 Sep 2019 23:15:06 GMT]
   Server: [Windows-Azure-File/1.0 Microsoft-HTTPAPI/2.0]
   X-Ms-Error-Code: [ShareBeingDeleted]
   X-Ms-Request-Id: [58c3b74a-f01a-0185-5fbf-69ba2e000000]
   X-Ms-Version: [2018-03-28]
[...]

PR #298, opened 2020-06-01
fix: switch to use management sdk

### LINKED ISSUE #188: Azure Files PV AuthorizationFailure when using advanced networking
**Is your feature request related to a problem? Please describe.**

PR(https://github.com/kubernetes/kubernetes/pull/85475) said it fixed the [Azure Files PV AuthorizationFailure when using advanced networking](https://github.com/kubernetes/kubernetes/issues/85354) issue on 1.18.0, actually it does not.

By creating an azure file share, it's using `github.com/Azure/azure-sdk-for-go/storage`:
https://github.com/kubernetes/kubernetes/blob/672aa55ee4860a8ae497c0523bbaf4ab9ac169a0/staging/src/k8s.io/legacy-cloud-providers/azure/azure_file.go#L87

while only following sdk should fix this issue:
https://github.com/Azure/azure-sdk-for-go/blob/master/services/storage/mgmt/2019-06-01/storage/fileshares.go

So we should use switch to use 2019-06-01 file share sdk
@VybavaRamadoss @RenaShahMSFT

BTW, using azure file csi driver would wokaround this [Azure Files PV AuthorizationFailure when using advanced networking](https://github.com/kubernetes/kubernetes/issues/85354) issue now since the azure file  csi driver would run on agent node(within vnet) which means it's not restricted, while for built-in tree, it runs on AKS controller node which is out of vnet.

**Additional context**

ZeroMagic:
**What type of PR is this?**
/kind feature
/kind bug

- switch to use management sdk
- handle ShareBeingDeleted error 
- handle the problem that PV is stuck at terminating after PVC is deleted during e2e tests

Fixes #188 #311 #119 

In `func ControllerExpandVolume()`, we could not get the file share now. So the current quota could not be printed. Later we should add the func Get() in k/k.

**Release note**:
```
switch to use management sdk
```

### DIFF
--- a/pkg/azurefile/controllerserver.go
+++ b/pkg/azurefile/controllerserver.go
@@ -148,6 +148,10 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 				klog.Warningf("CreateFileShare(%s) on account(%s) failed with error(%v), sleep 3s to retry", validFileShareName, account, retErr)
 				time.Sleep(3 * time.Second)
 				return false, nil
+			} else if strings.Contains(retErr.Error(), shareBeingDeleted) {
+				klog.Warningf("CreateFileShare(%s) on account(%s) failed with error(%v), sleep 1s to retry", validFileShareName, account, retErr)
+				time.Sleep(time.Second)
+				return false, nil
 			}
 		}
 		return true, retErr
--- a/pkg/azurefile/controllerserver.go
+++ b/pkg/azurefile/controllerserver.go
@@ -40,7 +40,7 @@ import (
 
 // CreateVolume provisions an azure file
 func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest) (*csi.CreateVolumeResponse, error) {
-	klog.V(2).Infof("CreateVolume called with request %v", *req)
+	klog.V(2).Infof("CreateVolume called with request %+v", *req)
 	if err := d.ValidateControllerServiceRequest(csi.ControllerServiceCapability_RPC_CREATE_DELETE_VOLUME); err != nil {
 		klog.Errorf("invalid create volume req: %v", req)
 		return nil, err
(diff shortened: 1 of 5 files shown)


andyzhangx:
I think you can update azure cloud provider in first PR, and then do the code change in driver in second PR.

andyzhangx:
could you upgrade go sdk to absorb this PR: https://github.com/kubernetes/kubernetes/pull/91861, may need one day or two when this PR merged into k/k

ZeroMagic:
> could you upgrade go sdk to absorb this PR: [kubernetes/kubernetes#91861](https://github.com/kubernetes/kubernetes/pull/91861), may need one day or two when this PR merged into k/k

Sure. I will do it in another PR firstly. Then update this PR.

ZeroMagic:
Because [k8s-publishing-bot](https://github.com/k8s-publishing-bot) broke currently, we need to wait for the fix. After that, the kubernetes/kubernetes#91861 will be merged.

pkg/azurefile/controllerserver.go
```diff
@@ -136,6 +148,10 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 				klog.Warningf("CreateFileShare(%s) on account(%s) failed with error(%v), sleep 3s to retry", validFileShareName, account, retErr)
 				time.Sleep(3 * time.Second)
 				return false, nil
+			} else if strings.Contains(retErr.Error(), shareBeingDeleted) {
```
andyzhangx, line 151: need to check whether we could check return code instead of error msg later

andyzhangx: looks like it only returns error msg:
https://github.com/Azure/azure-sdk-for-go/blob/4f14e3d53fba12172734d383af647730d06412d2/services/storage/mgmt/2019-06-01/storage/fileshares.go#L56

ZeroMagic: Yep. All apis in azure-sdk-for-go just return error msg.

PR #316, opened 2020-06-23
test: Add tests for csicommon/server

animeshk08:
Signed-off-by: Animesh Kumar <animuz111@gmail.com>

**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespaces from that line:
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation

> /kind failing-test
> /kind feature
> /kind flake

/kind test

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [x] adds unit tests

**Release note**:
```
test: Add tests for csicommon/server
```

### DIFF
--- a/pkg/csi-common/server.go
+++ b/pkg/csi-common/server.go
@@ -17,21 +17,20 @@ limitations under the License.
 package csicommon
 
 import (
+	"github.com/container-storage-interface/spec/lib/go/csi"
+	"google.golang.org/grpc"
+	"k8s.io/klog"
 	"net"
 	"os"
 	"runtime"
 	"sync"
-
-	"google.golang.org/grpc"
-	"k8s.io/klog"
-
-	"github.com/container-storage-interface/spec/lib/go/csi"
+	"time"
 )
 
 // Defines Non blocking GRPC server interfaces
 type NonBlockingGRPCServer interface {
 	// Start services at the endpoint
-	Start(endpoint string, ids csi.IdentityServer, cs csi.ControllerServer, ns csi.NodeServer)
+	Start(endpoint string, ids csi.IdentityServer, cs csi.ControllerServer, ns csi.NodeServer, testMode bool)
 	// Waits for the service to stop
 	Wait()
 	// Stops the service gracefully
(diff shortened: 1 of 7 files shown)


animeshk08:
This was a very tricky UT to write.
*  As discussed the server runs continuously hence, the unit tests never terminated. To prevent this a new parameter `testMode` has been added. This is then used to create a go-routine to close the server in case the methods are run in the `testMode`. This is the simplest way to cover these lines without doing any major change in the source code. Some other alternatives could have been to expose the server out of the `Run` method even then racing condition was introduced.

* To prevent the racing condition the logic for stopping the server is included within the method `serve` rather than including it inside specific unit-tests.

* A lot of the lines still remain uncovered. These primarily include ones involving `klog.fatalf` statement. This statement makes the program to exit with status code 255. Thus the UT also exits. On doing some digging I was able to find a [work around](https://blog.antoine-augusti.fr/2015/12/testing-an-os-exit-scenario-in-golang/). This required creating a subprocess, thus even though the UT ran and assertions could be tested ut coverage could not be generated. I also looked through upstream Kubernetes repo and was not able to find any examples to cover such cases. Please let me know in case one is found.

animeshk08:
Partially fixes #258

animeshk08:
The PR is ready for review @andyzhangx. I have tested it locally and the server is closed only while running in `testMode`

andyzhangx:
thanks. @animeshk08 I think it's ok now for writing such UT for csicommon/server, dig into more details on point# 3 may not help a lot. Could you work on other UTs? thanks.

@ZeroMagic could you also help review whether there is issue? We may apply such unit tests for all the other CSI drivers.
cc @Sakuralbj

ZeroMagic:
It looks good to me

pkg/csi-common/server.go
```diff
@@ -50,9 +49,11 @@ type nonBlockingGRPCServer struct {
 	server *grpc.Server
 }
 
-func (s *nonBlockingGRPCServer) Start(endpoint string, ids csi.IdentityServer, cs csi.ControllerServer, ns csi.NodeServer) {
+//var TestBool = false
```
andyzhangx, line 52: nit: remove this in next PR, thanks.

pkg/csi-common/server.go
```diff
@@ -103,6 +104,17 @@ func (s *nonBlockingGRPCServer) serve(endpoint string, ids csi.IdentityServer, c
 	if ns != nil {
 		csi.RegisterNodeServer(server, ns)
 	}
+	// Used to stop the server while running tests
+	if testMode {
+		s.wg.Done()
+		go func() {
+			// make sure Serve() is called
+
```
andyzhangx, line 112: nit: remove empty line in next PR.

pkg/csi-common/server.go
animeshk08: My bad. Will do it.

pkg/csi-common/server.go
```diff
@@ -17,21 +17,20 @@ limitations under the License.
 package csicommon
 
 import (
+	"github.com/container-storage-interface/spec/lib/go/csi"
```
andyzhangx, line 20: nit: pls keep sequence of import packages, e.g. first basic package, then github packages.

PR #340, opened 2020-07-10
fix: retry on EnsureStorageAccount if it is throttled

### LINKED ISSUE #323: retry on EnsureStorageAccount if it's throttled

retry on EnsureStorageAccount if it's throttled, now it would  return error directly:

https://github.com/kubernetes-sigs/azurefile-csi-driver/blob/5d9d6fc69bf849e96b3c0850516eb518451bbc10/pkg/azurefile/controllerserver.go#L122-L125

```
I0627 12:41:38.068974    8448 controllerserver.go:43] CreateVolume called with request {Name:sanity-controller-unpublish-CE796455-C72C37D5 CapacityRange:<nil> VolumeCapabilities:[mount:<> access_mode:<mode:SINGLE_NODE_WRITER > ] Parameters:map[fstype:ext4] Secrets:map[] VolumeContentSource:<nil> AccessibilityRequirements:<nil> XXX_NoUnkeyedLiteral:{} XXX_unrecognized:[] XXX_sizecache:0}
W0627 12:41:38.069112    8448 controllerserver.go:61] no quota specified, set as default value(100 GiB)
E0627 12:41:38.069160    8448 utils.go:116] GRPC error: rpc error: code = Internal desc = failed to ensure storage account: could not list storage accounts for account type : Retriable: true, RetryAfter: 16s, HTTPStatusCode: 0, RawError: azure cloud provider throttled for operation StorageAccountListByResourceGroup with reason "client throttled"
```

**Additional context**

ZeroMagic:
**What type of PR is this?**
/kind bug

Retry on EnsureStorageAccount if it's throttled.

Fixes #323

**Requirements**:
  
- [ ] includes documentation
- [x] adds unit tests

**Release note**:
```
retry on EnsureStorageAccount if it's throttled
```

### DIFF
--- a/pkg/azurefile/controllerserver.go
+++ b/pkg/azurefile/controllerserver.go
@@ -21,7 +21,6 @@ import (
 	"fmt"
 	"net/url"
 	"strings"
-	"time"
 
 	volumehelper "sigs.k8s.io/azurefile-csi-driver/pkg/util"
 
@@ -133,7 +132,18 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 	}
 
 	enableHTTPSTrafficOnly := true
-	retAccount, retAccountKey, err := d.cloud.EnsureStorageAccount(account, sku, accountKind, resourceGroup, location, fileShareAccountNamePrefix, enableHTTPSTrafficOnly)
+
+	var retAccount, retAccountKey string
+	err := wait.ExponentialBackoff(d.cloud.RequestBackoff(), func() (bool, error) {
+		var retErr error
+		retAccount, retAccountKey, retErr = d.cloud.EnsureStorageAccount(account, sku, accountKind, resourceGroup, location, fileShareAccountNamePrefix, enableHTTPSTrafficOnly)
+		if isRetriableError(retErr) {
+			klog.Warningf("EnsureStorageAccount(%s) failed with error(%v), waiting for retring", account, retErr)
+			return false, nil
+		}
+		return true, retErr
+	})
+
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, "failed to ensure storage account: %v", err)
 	}
@@ -146,28 +156,12 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 	lockKey := retAccount + sku + accountKind + resourceGroup + location
 	d.volLockMap.LockEntry(lockKey)
 	defer d.volLockMap.UnlockEntry(lockKey)
-	err = wait.PollImmediate(1*time.Second, 2*time.Minute, func() (bool, error) {
+	err = wait.ExponentialBackoff(d.cloud.RequestBackoff(), func() (bool, error) {
 		var retErr error
 		retAccount, retAccountKey, retErr = d.cloud.CreateFileShare(validFileShareName, account, sku, accountKind, resourceGroup, location, protocol, fileShareSize)
-		if retErr != nil {
-			if strings.Contains(retErr.Error(), accountNotProvisioned) {
-				klog.Warningf("CreateFileShare(%s) on account(%s) failed with error(%v), sleep 1s to retry", validFileShareName, account, retErr)
-				time.Sleep(time.Second)
-				return false, nil
-			} else if strings.Contains(retErr.Error(), tooManyRequests) {
-				// this is a workaround fix for 429 throttling issue, will update cloud provider for better fix later
-				klog.Warningf("CreateFileShare(%s) on account(%s) failed with error(%v), sleep 10s to retry", validFileShareName, account, retErr)
-				time.Sleep(10 * time.Second)
-				return false, nil
-			} else if strings.Contains(retErr.Error(), shareNotFound) {
-				klog.Warningf("CreateFileShare(%s) on account(%s) failed with error(%v), sleep 3s to retry", validFileShareName, account, retErr)
-				time.Sleep(3 * time.Second)
-				return false, nil
-			} else if strings.Contains(retErr.Error(), shareBeingDeleted) {
-				klog.Warningf("CreateFileShare(%s) on account(%s) failed with error(%v), sleep 1s to retry", validFileShareName, account, retErr)
-				time.Sleep(time.Second)
-				return false, nil
-			}
+		if isRetriableError(retErr) {
+			klog.Warningf("CreateFileShare(%s) on account(%s) failed with error(%v), waiting for retring", validFileShareName, account, retErr)
+			return false, nil
 		}
 		return true, retErr
 	})
--- a/pkg/azurefile/utils.go
+++ b/pkg/azurefile/utils.go
@@ -19,6 +19,7 @@ package azurefile
 import (
 	"context"
 	"fmt"
+	"strings"
 	"sync"
 
 	v1 "k8s.io/api/core/v1"
@@ -125,6 +126,17 @@ func isSupportedFsType(fsType string) bool {
 	for _, v := range supportedFsTypeList {
 		if fsType == v {
 			return true
+		}
+	}
+	return false
+}
+
+func isRetriableError(err error) bool {
+	if err != nil {
+		for _, v := range retriableErrors {
+			if strings.Contains(strings.ToLower(err.Error()), strings.ToLower(v)) {
+				return true
+			}
 		}
 	}
 	return false
--- a/pkg/azurefile/utils_test.go
+++ b/pkg/azurefile/utils_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package azurefile
 
 import (
+	"errors"
 	"fmt"
 	"reflect"
 	"testing"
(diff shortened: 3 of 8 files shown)


pkg/azurefile/utils.go
```diff
@@ -129,3 +131,21 @@ func isSupportedFsType(fsType string) bool {
 	}
 	return false
 }
+
+func checkAzureRPCError(shareName, accountName string, err error) (bool, error) {
+	if err != nil {
+		for k, v := range sleepTimeWithAzureRPCErrors {
+			if strings.Contains(err.Error(), k) {
+				if k == clientThrottled {
+					klog.Warningf("EnsureStorageAccount(%s) failed with error(%v), sleep %ds to retry", accountName, err, v)
```
andyzhangx, line 140: in this fuction, actually you don't know whether it's `EnsureStorageAccount` or `CreateFileShare`, what about print out message outside of this function?

pkg/azurefile/utils_test.go
```diff
@@ -243,3 +244,71 @@ func TestIsSupportedFsType(t *testing.T) {
...
+			accountName:  "",
+			rpcErr:       nil,
+			expectedBool: true,
+			expectedErr:  nil,
+		},
+		{
+			desc:         "fail with clientThrottled",
+			shareName:    "test-share",
+			accountName:  "",
+			rpcErr:       errors.New("could not list storage accounts for account type : Retriable: true, RetryAfter: 16s, HTTPStatusCode: 0, RawError: azure cloud provider throttled for operation StorageAccountListByResourceGroup with reason \"client throttled\""),
```
andyzhangx, line 269: it already said `RetryAfter: 16s`

ZeroMagic: done

pkg/azurefile/utils.go
ZeroMagic: done.

ZeroMagic:
switched to use `wait.ExponentialBackoff(d.cloud.RequestBackoff(), func() (bool, error) {})`

ZeroMagic:
Default `cloudProviderBackoffRetries` is 6, and default `cloudProviderBackoffDuration` is 5 in `azure.json`. Then the max timeout is the 30s. In some scenarios, it is not enough to wait for creating file share successfully. So the sanity ci test failed. 

It seems that there is not a good way to change the values. Do we still change to use `azure_backoff`?

andyzhangx:
> Default `cloudProviderBackoffRetries` is 6 in `azure.json`. In some scenarios, the max retry is not enough to wait for creating file share successfully. So the sanity ci test failed.
> 
> It seems that there is not a good way to change the values. Do we still change to use `azure_backoff`?

then don't use the value from `azure.json` and  use the backoff function?

ZeroMagic:
`https://github.com/kubernetes/kubernetes/blob/2d327ac4558d78c744004db178dacb80bd6e0b9e/staging/src/k8s.io/legacy-cloud-providers/azure/azure_backoff.go#L47-L59`

The function still uses the cloud fields which is read from `azure.json`

pkg/azurefile/controllerserver.go
```diff
@@ -133,7 +133,20 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 	}
 
 	enableHTTPSTrafficOnly := true
-	retAccount, retAccountKey, err := d.cloud.EnsureStorageAccount(account, sku, accountKind, resourceGroup, location, fileShareAccountNamePrefix, enableHTTPSTrafficOnly)
+
+	var retAccount, retAccountKey string
+	err := wait.ExponentialBackoff(d.cloud.RequestBackoff(), func() (bool, error) {
```
andyzhangx, line 138: the usage of this is incorrect, don't need to sleep in inner function, return value is incorrect.
```
// ExponentialBackoff repeats a condition check with exponential backoff.
//
// It repeatedly checks the condition and then sleeps, using `backoff.Step()`
// to determine the length of the sleep and adjust Duration and Steps.
// Stops and returns as soon as:
// 1. the condition check returns true or an error,
// 2. `backoff.Steps` checks of the condition have been done, or
// 3. a sleep truncated by the cap on duration has been completed.
// In case (1) the returned error is what the condition function returned.
// In all other cases, ErrWaitTimeout is returned.
func ExponentialBackoff(backoff Backoff, condition ConditionFunc) error {
	for backoff.Steps > 0 {
		if ok, err := runConditionWithCrashProtection(condition); err != nil || ok {
			return err
		}
		if backoff.Steps == 1 {
			break
		}
		time.Sleep(backoff.Step())
	}
	return ErrWaitTimeout
}
```

andyzhangx:
the `DumpClusterLogs` failure is expected, there is an ongoing PR to fix this issue

pkg/azurefile/utils.go
```diff
@@ -129,3 +130,14 @@ func isSupportedFsType(fsType string) bool {
 	}
 	return false
 }
+
+func isUnkownAzureRPCError(err error) bool {
```
andyzhangx, line 134: isRetriableError ?

pkg/azurefile/utils.go
```diff
@@ -129,3 +130,14 @@ func isSupportedFsType(fsType string) bool {
...
+	if err != nil {
+		for _, v := range knownAzureRPCErrors {
+			if strings.Contains(err.Error(), v) {
```
andyzhangx, line 137: do we have a case in-sensitive contain function?

pkg/azurefile/controllerserver.go
```diff
@@ -133,7 +133,20 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
...
+		var retErr error
+		retAccount, retAccountKey, retErr = d.cloud.EnsureStorageAccount(account, sku, accountKind, resourceGroup, location, fileShareAccountNamePrefix, enableHTTPSTrafficOnly)
+		isRetriable, sleepTime := isRetriableError(retErr)
+		if isRetriable {
+			klog.Warningf("EnsureStorageAccount(%s) failed with error(%v), sleep %ds to retry", account, retErr, sleepTime)
+			time.Sleep(time.Duration(sleepTime) * time.Second)
```
andyzhangx, line 144: there is already sleep in `ExponentialBackoff`, is it necessary here?

pkg/azurefile/controllerserver.go
```diff
@@ -146,28 +159,14 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
...
-				return false, nil
-			} else if strings.Contains(retErr.Error(), shareBeingDeleted) {
-				klog.Warningf("CreateFileShare(%s) on account(%s) failed with error(%v), sleep 1s to retry", validFileShareName, account, retErr)
-				time.Sleep(time.Second)
-				return false, nil
-			}
+		isRetriable, sleepTime := isRetriableError(retErr)
+		if isRetriable {
+			klog.Warningf("CreateFileShare(%s) on account(%s) failed with error(%v), sleep %ds to retry", validFileShareName, account, retErr, sleepTime)
+			time.Sleep(time.Duration(sleepTime) * time.Second)
```
andyzhangx, line 168: there is already sleep in `ExponentialBackoff`, is it necessary here?

pkg/azurefile/controllerserver.go
ZeroMagic: I am not sure. Wait for the tests again.

ZeroMagic: According to the result, if we just sleep in ExponentialBackoff, sometimes the sanity tests still fail. 
Following is the bacakoff we set in sanity tests.
```
defaultCloudProviderBackoff         = true
defaultCloudProviderBackoffRetries  = 6
defaultCloudProviderBackoffDuration = 5
```

andyzhangx: try increase `defaultCloudProviderBackoffDuration` in the sanity test?

pkg/azurefile/controllerserver.go
```diff
@@ -133,7 +132,19 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 	}
 
 	enableHTTPSTrafficOnly := true
-	retAccount, retAccountKey, err := d.cloud.EnsureStorageAccount(account, sku, accountKind, resourceGroup, location, fileShareAccountNamePrefix, enableHTTPSTrafficOnly)
+
+	var retAccount, retAccountKey string
+	err := wait.ExponentialBackoff(d.cloud.RequestBackoff(), func() (bool, error) {
+		var retErr error
+		retAccount, retAccountKey, retErr = d.cloud.EnsureStorageAccount(account, sku, accountKind, resourceGroup, location, fileShareAccountNamePrefix, enableHTTPSTrafficOnly)
+		isRetriable := isRetriableError(retErr)
```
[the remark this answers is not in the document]
ZeroMagic: done

ZeroMagic:
Let's merge this PR firstly. I will tune the backoff config in another PR.

PR #372, opened 2020-08-14
Add value in Helm chart for secret used to pull Docker images

### LINKED ISSUE #371: Helm chart should allow secrets to pull Docker images
The Helm chart allow to change Docker images repository when they are not hosted on Docker hub. But it does not allow to provide a secret used to authenticate in case these repositories are private.

The chart should add a optional variable thats contains secrets used to authenticate. These secrets could then be attached to the pods to pull the images.

I can provide the pull request.

fredgate:
/kind feature

Allow to specify a secret in Helm chart used to provide credentials for Docker repositories (when they are private).

Fixes #371 

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

### DIFF
--- a/charts/latest/azurefile-csi-driver/values.yaml
+++ b/charts/latest/azurefile-csi-driver/values.yaml
@@ -23,6 +23,9 @@ image:
     repository: mcr.microsoft.com/oss/kubernetes-csi/csi-node-driver-registrar
     tag: v1.2.0
     pullPolicy: IfNotPresent
+## Name of secrets (present in namespace) containing Docker credentials used to pull images
+# imagePullSecrets:
+#   - myRegistryKeySecretName
 
 serviceAccount:
   create: true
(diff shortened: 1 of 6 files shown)


andyzhangx:
@fredgate so in your case, all csi driver images are in private docker repo? Are those agent nodes running on Azure?

fredgate:
@andyzhangx yes all these images can be in a private Docker registry, when our customer does not allow access to Internet.
These agent nodes are running on Azure (in a AKS).

charts/latest/azurefile-csi-driver/values.yaml
```diff
@@ -23,6 +23,9 @@ image:
     repository: mcr.microsoft.com/oss/kubernetes-csi/csi-node-driver-registrar
     tag: v1.2.0
     pullPolicy: IfNotPresent
+## Name of secrets (present in namespace) containing Docker credentials used to pull images
```
andyzhangx, line 26: this is where we put default values, so it's not necessary to add comments here since by default it's empty

andyzhangx:
pls skip the windows test failure since windows image is lost, that's an incident ...

charts/latest/azurefile-csi-driver/values.yaml
andyzhangx: @fredgate are you going to fix this? just remove the comments is ok

andyzhangx (approved):
/lgtm
/approve
let's merge this first

PR #384, opened 2020-08-27
chore: remove unnecessary check

andyzhangx:
**What type of PR is this?**
/kind cleanup

chore: remove unnecessary check

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/azurefile/azurefile.go
```diff
@@ -429,11 +429,7 @@ func (d *Driver) GetAccountInfo(volumeID string, secrets, reqContext map[string]
 				diskName = v
 			}
 		}
-		if fileShareName != "" {
-			accountName, accountKey, err = getStorageAccount(secrets)
-		} else {
-			err = fmt.Errorf("could not find sharename from context(%v)", reqContext)
-		}
+		accountName, accountKey, err = getStorageAccount(secrets)
```
ZeroMagic, line 432: Why isn't the judgement required? If the fileShareName isn't specified, the return value is "". Then we can not get the fileURL in `ControllerPublishVolume` and `ControllerUnpublishVolume`.

andyzhangx: we want to make `getStorageAccount` as simple as possible, and if `fileShareName` is required, then the invoker should check whether `fileShareName`. And for this case, `ControllerPublishVolume` and `ControllerUnpublishVolume` does not need `fileShareName`

ZeroMagic (approved):
/lgtm

PR #454, opened 2020-10-11
Fix tolerations in the daemon-set.

monaka:
**What type of PR is this?**

Stops deploying the daemon-set pod to virtual-kubelet.

/kind bug

AFAIK, virtual-kubelet on Azure doesn't support CSI.

None.

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/deploy/csi-azurefile-node.yaml
+++ b/deploy/csi-azurefile-node.yaml
@@ -19,8 +19,6 @@ spec:
       nodeSelector:
         kubernetes.io/os: linux
       priorityClassName: system-node-critical
-      tolerations:
-        - operator: "Exists"
       containers:
         - name: liveness-probe
           volumeMounts:


deploy/csi-azurefile-node.yaml
```diff
@@ -20,7 +20,10 @@ spec:
         kubernetes.io/os: linux
       priorityClassName: system-node-critical
       tolerations:
-        - operator: "Exists"
+        - effect: NoSchedule
+          key: type
+          operator: Equal
```
andyzhangx, line 25: @monaka thanks for the fix. This fix actually means daemonset pod would be scheduled to `virtual-kubelet` node per https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/#concepts

andyzhangx: simply removes `operator: "Exists"` could work, while this driver daemonset would not run on one agent node with taint. This requirement for this PR is actually this:
 - ignore all taints except `virtual-kubelet`
Looks like there is this is not supported in k8s taint functionality.

andyzhangx: question: is there any side effect if this daemonset pod is scheduled to virtual-kubelet? `operator: "Exists"` will create a new ACI instance which would make user pay more cost?

monaka: `effect: NoSchedule` means to avoid deploying pods to nodes tagged as `type=virtual-kubelect`.

andyzhangx: that's opposite, check 
https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/#concepts
```
A toleration "matches" a taint if the keys are the same and the effects are the same, and
```

monaka:
I see both of my thought and the code bad.
This issue is fixed just by removing `toleration:`.
I'm going to push new patch here.

andyzhangx:
> toleration

removing `toleration` may bring other issue, what's the side effect if we don't fix it?

monaka:
What is "other issue" ?  Strictly, I can agree all operations in the world will cause "other issue" anyway.

Current code have an obvious issue. it causes a deployment failure if virtual-kubelet exists in the cluster. Can you agree?

andyzhangx:
> What is "other issue" ? Strictly, I can agree all operations in the world will cause "other issue" anyway.
> 
> Current code have an obvious issue. it causes a deployment failure if virtual-kubelet exists in the cluster. Can you agree?

azure file & azure disk are basic drivers that should run everywhere, there is case that use don't want to schedule normal pods on some nodes, e.g. GPU nodes:
```console
kubectl taint node aks-nodepool1 sku=gpu:NoSchedule
```

As long as user set taints, azure file & azure disk daemonset pods could not run on those nodes with taints.

andyzhangx:
I have deployed this driver on a ACI enabled cluster, I don't find there is a driver daemonset pod running on ACI:

```
# k get po -n kube-system -o wide | grep csi
csi-azurefile-controller-7c47787bb7-mxstt    6/6     Running   0          50s   10.240.0.115   aks-agentpool-20968559-vmss000001   <none>           <none>
csi-azurefile-controller-7c47787bb7-nm8fp    6/6     Running   0          50s   10.240.0.4     aks-agentpool-20968559-vmss000000   <none>           <none>
csi-azurefile-node-sfqfk                     3/3     Running   0          50s   10.240.0.115   aks-agentpool-20968559-vmss000001   <none>           <none>
csi-azurefile-node-wpjwn                     3/3     Running   0          50s   10.240.0.4     aks-agentpool-20968559-vmss000000   <none>           <none>
# k get no -o wide
NAME                                STATUS   ROLES   AGE   VERSION                         INTERNAL-IP    EXTERNAL-IP   OS-IMAGE             KERNEL-VERSION      CONTAINER-RUNTIME
aks-agentpool-20968559-vmss000000   Ready    agent   15m   v1.17.11                        10.240.0.4     <none>        Ubuntu 16.04.7 LTS   4.15.0-1096-azure   docker://19.3.12
aks-agentpool-20968559-vmss000001   Ready    agent   15m   v1.17.11                        10.240.0.115   <none>        Ubuntu 16.04.7 LTS   4.15.0-1096-azure   docker://19.3.12
virtual-node-aci-linux              Ready    agent   15m   v1.14.3-vk-azure-aci-v1.2.1.1   10.240.0.126   <none>        <unknown>            <unknown>           <unknown>
```

andyzhangx:
another way is use anti-affinity since ACI node has different label, while I am not sure where is broken on ACI, try: https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#node-affinity

andyzhangx:
I think this PR could fix the scheduling issue on virtual kubelet more gracefully: https://github.com/kubernetes-sigs/azurefile-csi-driver/pull/461 @monaka PTAL, thanks

andyzhangx:
close this PR since it's already fixed by https://github.com/kubernetes-sigs/azurefile-csi-driver/pull/461

PR #463, opened 2020-10-29
Add AZURE prefix to e2e environment variables

### LINKED ISSUE #459: Consider adding AZURE prefix to all env variables in E2E
Running tests in environments which deal with multiple providers, not just Azure, is a pretty common use case. We should consider renaming all the Azure specific environment variables to add an `AZURE_` prefix such that there are no collisions with other providers. For example, `LOCATION` should be `AZURE_LOCATION`. This might require some changes in kubetest as well. 

This is a follow up from https://github.com/kubernetes/test-infra/pull/19699.

/cc @chewong

CecileRobertMichon:
**What type of PR is this?**
/kind cleanup

Fixes #459 

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

/hold

I need to go through all the repos that rely on these values.

**Release note**:
```
Add AZURE prefix to e2e environment variables
```

### DIFF
--- a/test/utils/credentials/credentials.go
+++ b/test/utils/credentials/credentials.go
@@ -54,19 +54,19 @@ const (
 	defaultCloudProviderBackoffDuration = 5
 
 	// Env vars
-	tenantIDEnvVar        = "TENANT_ID"
-	subscriptionIDEnvVar  = "SUBSCRIPTION_ID"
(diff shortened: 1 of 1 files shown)


andyzhangx (approved):
/lgtm
/approve
/hold cancel

Issue #482, opened 2020-12-02
Azurefile should validate volume capabilities


Jiawei0227:
I am running K8s subpath testsuite against azurefile driver. And this test failed:

```
[Fail] External Storage [Driver: file.csi.azure.com] [Testpattern: Dynamic PV (block volmode)] 
volumeMode [It] should fail in binding dynamic provisioned PV to PVC [Slow][LinuxOnly]
xx/go/src/k8s.io/kubernetes/test/e2e/storage/testsuites/volumemode.go:285
```

So it seems like if user specific a block volume mode in the volume capability. We did not check if that is valid or not.

/cc @msau42 
/cc @andyzhangx

PR #483, opened 2020-12-02
fix: return error for creating raw block volume(not supported)

Jiawei0227:
**What type of PR is this?**
/kind bug

This fixes an issue that when we call the CreateVolume with block volume capability, the controller service should validate that and return an error.

Fixes # https://github.com/kubernetes-sigs/azurefile-csi-driver/issues/482

**Release note**:
```
FIX: Return error for creating raw block volume(not supported)
```

/cc @msau42 
/assign @andyzhangx

### DIFF
--- a/pkg/azurefile/controllerserver.go
+++ b/pkg/azurefile/controllerserver.go
@@ -44,6 +44,26 @@ const (
 	azureFileCSIDriverName = "azurefile_csi_driver"
 )
 
+var (
+	volumeCaps = []csi.VolumeCapability_AccessMode{
+		{
+			Mode: csi.VolumeCapability_AccessMode_SINGLE_NODE_WRITER,
+		},
+		{
+			Mode: csi.VolumeCapability_AccessMode_SINGLE_NODE_READER_ONLY,
+		},
+		{
+			Mode: csi.VolumeCapability_AccessMode_MULTI_NODE_READER_ONLY,
+		},
+		{
+			Mode: csi.VolumeCapability_AccessMode_MULTI_NODE_SINGLE_WRITER,
+		},
+		{
+			Mode: csi.VolumeCapability_AccessMode_MULTI_NODE_MULTI_WRITER,
+		},
+	}
+)
+
 // CreateVolume provisions an azure file
 func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest) (*csi.CreateVolumeResponse, error) {
 	if err := d.ValidateControllerServiceRequest(csi.ControllerServiceCapability_RPC_CREATE_DELETE_VOLUME); err != nil {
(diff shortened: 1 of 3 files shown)


pkg/azurefile/controllerserver.go
```diff
@@ -704,3 +724,28 @@ func (d *Driver) snapshotExists(ctx context.Context, sourceVolumeID, snapshotNam
...
 	return false, azfile.ShareItem{}, nil
 }
+
+// isValidVolumeCapabilities validates the given VolumeCapability array is valid
+func isValidVolumeCapabilities(volCaps []*csi.VolumeCapability) error {
+	if len(volCaps) == 0 {
+		return fmt.Errorf("CreateVolume Volume capabilities must be provided")
+	}
+	hasSupport := func(cap *csi.VolumeCapability) error {
+		if blk := cap.GetBlock(); blk != nil {
```
andyzhangx, line 734: does this mean Azure File driver does not support block volume, right?

andyzhangx: I think Azure File should not support block volume, block volume only works for disk

Jiawei0227: To my understanding it should not support raw block volume. But correct me if I am wrong.

test/integration/run-test.sh
```diff
@@ -62,22 +62,22 @@ fi
 
 # Begin to run CSI functions one by one
 echo 'Create volume test:'
-readonly value=$("$CSC_BIN" controller new --endpoint "$endpoint" --cap 1,block "$volname" --req-bytes 2147483648 --params "$params")
```
Jiawei0227, line 65: I am not sure why these integration use block access_type. @andyzhangx I am changing this to mount. But please feel free to raise any concern if there is any.

andyzhangx: thanks, that's correct fix.

pkg/azurefile/controllerserver.go
andyzhangx: code part LGTM, thanks.
so the PR title could be `fix: return error for creating raw block volume(not supported)` ?

Jiawei0227: Done. Thanks

andyzhangx (approved):
/lgtm
/approve

msau42:
@andyzhangx does this impact the vhd feature?

Jiawei0227:
Does this three windows test using block volume or something? I am not sure why are they failing consitently.

andyzhangx:
> Does this three windows test using block volume or something? I am not sure why are they failing consistently.

it said ` {"error":{"code":"LocationNotValid","message":"Supplied location westus2 is not valid."}}}`, I think it's a transient Azure issue.

andyzhangx:
> @andyzhangx does this impact the vhd feature?

@msau42 I think vhd feature could not support raw block volume, there is no such requirement at least

andyzhangx:
still same error: `:{"code":"LocationNotValid","message":"Supplied location westus2 is not valid."}}}`, sth. wrong with the Azure subscription. Try run other tests too.

andyzhangx:
let me switch this test to eastus2, looks like `westus2` could not be used with this subs any more.

Issue #485, opened 2020-12-03
fsType in parameters can be confusing


Jiawei0227:
**What happened**:
Looking at the VHD support in Azurefile csi driver.

It seems user has to pass a fsType in the SC parameters to indicate this will be a VHD. This can be confusing because we also have `csi.storage.k8s.io/fstype` in the StorageClass.[1]

It could be changed to something like vhdFsType which might help in this case?

Also, I think right now Azurefile is reading fsType from req.Parameters, which seems problematic? Ideally, it should read this field from the VolumeCapability. I think here the CSI has some poor design but at least in csi-provisioner all the fsType in VolumeCapability is the same.

[1] https://kubernetes-csi.github.io/docs/external-provisioner.html#storageclass-parameters

Issue #484, opened 2020-12-03
set mfsymlinks as default mountOptions


andyzhangx:
**Is your feature request related to a problem?/Why is this needed**

**Describe the solution you'd like in detail**

set mfsymlinks as default mountOptions
Without this the symbolic operation will fail.

**Describe alternatives you've considered**

**Additional context**

andyzhangx:
fixed by https://github.com/kubernetes-sigs/azurefile-csi-driver/pull/487

PR #486, opened 2020-12-03
Change fsType to vhdFsType. Remove fsType in parameter support

Jiawei0227:
**What type of PR is this?**
/kind cleanup
/kind api-change

This PR remove fsType in the parameter supports as we also have 'csi.k8s.io/fstype' field in storageclass which can be confusing to users.
Instead, we rename it to vhdFsType.

Also, this PR remove `fsType: nfs` supports so whoever wants to use it need to set protocol instead of fsType.

Add mfsymlinks as the default mountOptions

Fixes # https://github.com/kubernetes-sigs/azurefile-csi-driver/issues/485 https://github.com/kubernetes-sigs/azurefile-csi-driver/issues/484

This is a breaking change so we might need to be more cautious about this! fsType in parameters will not be supported anymore.

**Release note**:
```
Rename fsType in parameters to vhdFsType. For filer fsType, Azurefile only supports cifs for now with different protocols(nfs, smb).
```


Jiawei0227:
This might only be compatible with csi-provisioner:v2.0 because you can specify default-fsType there...

Jiawei0227:
This is more of a POC. I think we need to at least rename it before GA

pkg/azurefile/controllerserver.go
```diff
@@ -90,8 +90,8 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 			fileShareName = v
 		case diskNameField:
 			diskName = v
-		case fsTypeField:
-			fsType = v
+		case vhdFsTypeField:
```
msau42, line 93: we may still need to support the original fstype parameter if backwards compatibility is important. Will let @andyzhangx decide.

pkg/azurefile/controllerserver.go
```diff
@@ -117,16 +121,13 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 	enableHTTPSTrafficOnly := true
 	shareProtocol := storage.SMB
 	var vnetResourceIDs []string
-	if fsType == nfs || protocol == nfs {
-		protocol = nfs
+	if protocol == nfs {
```
msau42, line 124: just for my own learning, when should nfs be used vs cifs?

pkg/azurefile/controllerserver.go
Jiawei0227: Yeah, I think we need to get rid of them before it goes to 1.0, but yeah... let's see

pkg/azurefile/controllerserver.go
Jiawei0227: @andyzhangx I also want to learn :)

andyzhangx: Azure file provides both SMB(by default) and NFS protocols

pkg/azurefile/controllerserver.go
andyzhangx: I think we still need to make backwards compatibility, is it possible to keep this `fsType`, e.g. we have two places to set this field: 1) `csi.k8s.io/fstype`  2) `fstype: ext4`  
```
if `csi.k8s.io/fstype` != "" {
  fsType = `csi.k8s.io/fstype` setting
} esle {
  fsType = `fstype` setting
}
```

Azure disk driver also has same issue, it's better not rename or remove existing parameters.

pkg/azurefile/controllerserver.go
Jiawei0227: Yeah, but is there a recommendation on when to use NFS over cifs? Normally how should user decides which one to use?

pkg/azurefile/controllerserver.go
Jiawei0227: Hmm, so I think at least one thing important is that we should read fsType from VolumeCapability instead of parameters.

And I feel like use fsType to decide when to use VHD over normal Azure file can be confusing. Would it possible to add a field vhdFsType and put that as the recommended field to set? And we can support fsType at the same time. But remove it once the driver wants to promote GA

Jiawei0227: And another reference:

Here is how gce-pd-csi-driver parses the fsType: 
https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/2f5ea47b054d091c9c97362f2e48721a7f8de698/pkg/gce-pd-csi-driver/node.go#L306

pkg/azurefile/controllerserver.go
andyzhangx: currently cifs is the default protocol

pkg/azurefile/controllerserver.go
andyzhangx: looks like we have 3 parameters now:
 - `parametes.fsType`
 - `parametes.vhdFsType`
 - `csi.k8s.io/fstype`

if we want to use `parametes.vhdFsType`, then had better not use `csi.k8s.io/fstype` since the naming is different, another option is only use:
 - `parametes.fsType`
 - `csi.k8s.io/fstype`

It's still open to discuss. Thanks.

msau42: Unfortunately because of https://github.com/kubernetes-csi/external-provisioner/issues/328, any PV's fstype is going to default to ext4 even if nothing was specified in the StorageClass.

That means existing PVs are going to have fstype ext4 by default, even if they were using cifs. So I think to retain backwards compatibility, we cannot use 'csi.k8s.io/fstype' to determine if something should be vhd or not. We need a specific parameter for vhd which then determines how we interpret the 'csi.k8s.io/fstype' field.

Jiawei0227: Hmm I think my point is that regardless how fsType are passing from parameters. The fsType should only be decided by VolumeCapabilities. And that is the way moving forward for CSI spec compatibility. Currently Azuredisk and Azurefile does not seem to read from VolumeCapabilities to get fsType.

andyzhangx: it's better that only add `vhdFsType` in this PR, and keep `fsType` as compatible, and due to this issue(https://github.com/kubernetes-csi/external-provisioner/issues/328), `csi.k8s.io/fstype` could be useless now since it's always `ext4` by default.

msau42: The external provisioner issue has been fixed in 2.0. However existing PVs are going to have PV.fstype set to ext4

PR #492, opened 2020-12-08
refactor: appendDefaultMountOptions to use map

### LINKED ISSUE #488: refactor appendDefaultMountOptions function

it's better use a map to store default MountOptions,  that would easier to add new default MountOptions in the future

https://github.com/kubernetes-sigs/azurefile-csi-driver/blob/3abdb132ee3fcd101337e11602e0c6e1d9182e90/pkg/azurefile/azurefile.go#L254-L277

**Additional context**

dr0pdb:
**What type of PR is this?**

/kind cleanup

This PR refactors the `appendDefaultMountOptions` function to use maps and slices for easier extensibility.

Fixes #488

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

N/A

**Release note**:
```
none
```

### DIFF
--- a/pkg/azurefile/azurefile.go
+++ b/pkg/azurefile/azurefile.go
@@ -253,10 +253,11 @@ func GetFileShareInfo(id string) (string, string, string, string, error) {
 // check whether mountOptions contains file_mode, dir_mode, vers, if not, append default mode
 func appendDefaultMountOptions(mountOptions []string) []string {
 	var defaultMountOptions = map[string]string{
-		fileMode: defaultFileMode,
-		dirMode:  defaultDirMode,
-		vers:     defaultVers,
-		actimeo:  defaultActimeo,
+		fileMode:   defaultFileMode,
+		dirMode:    defaultDirMode,
+		vers:       defaultVers,
+		actimeo:    defaultActimeo,
+		mfsymlinks: "",
 	}
 
 	// required for maintaing the order.
@@ -277,8 +278,8 @@ func appendDefaultMountOptions(mountOptions []string) []string {
 
 	for _, k := range keys {
 		if _, isIncluded := included[k]; !isIncluded {
-			if v, found := defaultMountOptions[k]; found {
-				allMountOptions = append(allMountOptions, fmt.Sprintf("%s=%s", k, v))
+			if defaultMountOptions[k] != "" {
+				allMountOptions = append(allMountOptions, fmt.Sprintf("%s=%s", k, defaultMountOptions[k]))
 			} else {
 				allMountOptions = append(allMountOptions, k)
 			}
--- a/pkg/azurefile/azurefile.go
+++ b/pkg/azurefile/azurefile.go
@@ -260,14 +260,11 @@ func appendDefaultMountOptions(mountOptions []string) []string {
 		mfsymlinks: "",
 	}
 
-	// required for maintaing the order.
-	var keys = []string{fileMode, dirMode, vers, actimeo, mfsymlinks}
-
 	// stores the mount options already included in mountOptions
 	included := make(map[string]bool)
 
 	for _, mountOption := range mountOptions {
-		for _, k := range keys {
+		for k, _ := range defaultMountOptions {
 			if strings.HasPrefix(mountOption, k) {
 				included[k] = true
 			}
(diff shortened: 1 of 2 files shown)


pkg/azurefile/azurefile.go
```diff
@@ -252,49 +252,37 @@ func GetFileShareInfo(id string) (string, string, string, string, error) {
...
+		vers:     defaultVers,
+		actimeo:  defaultActimeo,
 	}
 
-	allMountOptions := mountOptions
-	if !fileModeFlag {
-		allMountOptions = append(allMountOptions, fmt.Sprintf("%s=%s", fileMode, defaultFileMode))
-	}
+	// required for maintaing the order.
+	var keys = [5]string{fileMode, dirMode, vers, actimeo, mfsymlinks}
```
andyzhangx, line 263: why set `5` here? use a non-fixed array is better

andyzhangx (commented):
`mfsymlinks` is not a key-value pair, how did you handle that in this PR?

pkg/azurefile/azurefile.go
dr0pdb: Fixed

dr0pdb:
>mfsymlinks is not a key-value pair, how did you handle that in this PR?

The `defaultMountOptions` map only contains 4 key-value pairs. It doesn't contain `mfsymlinks`. So I check if a key is present in the map. If it's not then it's `mfsymlinks`.

```Go
if v, found := defaultMountOptions[k]; found {
    allMountOptions = append(allMountOptions, fmt.Sprintf("%s=%s", k, v))
} else {
    allMountOptions = append(allMountOptions, k)
}
```

andyzhangx:
> > mfsymlinks is not a key-value pair, how did you handle that in this PR?
> 
> The `defaultMountOptions` map only contains 4 key-value pairs while the `keys` array contains 5 keys. So if something is in the array but not in the map, then it is `mfsymlinks`.
> 
> ```go
> if v, found := defaultMountOptions[k]; found {
>     allMountOptions = append(allMountOptions, fmt.Sprintf("%s=%s", k, v))
> } else {
>     allMountOptions = append(allMountOptions, k)
> }
> ```

what about use one map <key, value>, and for `mfsymlinks`, it's <`mfsymlinks`, ``>, that would be more clear and not error prone.

dr0pdb:
>what about use one map <key, value>, and for mfsymlinks, it's <mfsymlinks, ``>, that would be more clear and not error prone.

Yes, but we still need if/else to handle the equals sign `=`. For the case of <mfsymlinks, ``>, it will add "mfsymlinks=" to the array. So an if/else is required for that also.

andyzhangx:
> > what about use one map <key, value>, and for mfsymlinks, it's <mfsymlinks, ``>, that would be more clear and not error prone.
> 
> Yes, but we still need if/else to handle the equals sign `=`. For the case of <mfsymlinks, ``>, it will add "mfsymlinks=" to the array. So an if/else is required for that also.

if value is empty, then don't need add `=`, I think that's more clear logic.

pkg/azurefile/azurefile.go
```diff
@@ -252,49 +252,38 @@ func GetFileShareInfo(id string) (string, string, string, string, error) {
...
+		actimeo:    defaultActimeo,
+		mfsymlinks: "",
 	}
 
-	allMountOptions := mountOptions
-	if !fileModeFlag {
-		allMountOptions = append(allMountOptions, fmt.Sprintf("%s=%s", fileMode, defaultFileMode))
-	}
+	// required for maintaing the order.
+	var keys = []string{fileMode, dirMode, vers, actimeo, mfsymlinks}
```
andyzhangx, line 264: why still need `keys`, keys is already in `defaultMountOptions`

dr0pdb: We need it to have a fixed iteration order. Map iteration is unordered and randomized. I thought we should keep a pre-defined order in the array. Otherwise, we can remove keys and then modify the unit tests.

andyzhangx: order is not important, feel free to modify the ut, thanks.

dr0pdb: Modified. Thanks.

andyzhangx (approved):
/lgtm
/approve

PR #513, opened 2021-01-02
fix: Handle concurrent operation of NodeUnstage with NodeStage and NodeUnpublish with NodePublish

### LINKED ISSUE #253: Handle concurrent operation of NodeUnstage with NodeStage and NodeUnpublish with NodePublish
**What happened**:
we don't correctly handle concurrent processing of NodeUnstage call if a previous NodeStage was in-progress and same thing applies to NodeUnpublish call too.

The specs says, if an operation is in-progress for the volume then SP should return error code "10 ABORTED" - https://github.com/container-storage-interface/spec/blob/master/spec.md#timeouts

/kind bug

**How to reproduce it**:

**Environment**:
- CSI Driver version:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

harshika-kashyap:
**What type of PR is this?**
/kind bug

we don't correctly handle concurrent processing of NodeUnstage call if a previous NodeStage was in-progress and same thing applies to NodeUnpublish call too.

The specs says, if an operation is in-progress for the volume then SP should return error code "10 ABORTED".

Fixes #253

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

```
none
```

**Release note**:
```
none
```

### DIFF
--- a/pkg/azurefile/nodeserver.go
+++ b/pkg/azurefile/nodeserver.go
@@ -56,6 +56,13 @@ func (d *Driver) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolu
 		return nil, status.Error(codes.InvalidArgument, "Staging target not provided")
 	}
 
+	volumeID := req.GetVolumeId()
+
+	if acquired := d.volumeLocks.TryAcquire(volumeID); !acquired {
+		return nil, status.Errorf(codes.Aborted, volumeOperationAlreadyExistsFmt, volumeID)
+	}
+	defer d.volumeLocks.Release(volumeID)
+
 	mountOptions := []string{"bind"}
 	if req.GetReadonly() {
 		mountOptions = append(mountOptions, "ro")
@@ -97,6 +104,11 @@ func (d *Driver) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpublish
 	}
 	targetPath := req.GetTargetPath()
 	volumeID := req.GetVolumeId()
+
+	if acquired := d.volumeLocks.TryAcquire(volumeID); !acquired {
+		return nil, status.Errorf(codes.Aborted, volumeOperationAlreadyExistsFmt, volumeID)
+	}
+	defer d.volumeLocks.Release(volumeID)
 
 	klog.V(2).Infof("NodeUnpublishVolume: unmounting volume %s on %s", volumeID, targetPath)
 	err := CleanupMountPoint(d.mounter, targetPath, false)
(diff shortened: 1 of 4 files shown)


harshika-kashyap:
CLA signed

pkg/azurefile/nodeserver.go
```diff
@@ -38,6 +38,11 @@ import (
 	"golang.org/x/net/context"
 )
 
+var (
+	volumeStageInProgress   map[string]bool
```
andyzhangx, line 42: map is not thread-safe, should use sync.Map

pkg/azurefile/nodeserver.go
```diff
@@ -56,6 +61,10 @@ func (d *Driver) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolu
 		return nil, status.Error(codes.InvalidArgument, "Staging target not provided")
 	}
 
+	volumeID := req.GetVolumeId()
+	volumePublishInProgress[volumeID] = true
+	defer delete(volumePublishInProgress, volumeID)
```
andyzhangx, line 66: what about there are two `NodePublishVolume` calls in same time, shall we handle that?

andyzhangx (commented):
it's better if we could add a unit test to test this scenario, thanks.

harshika-kashyap:
CLA signed

pkg/azurefile/nodeserver.go
andyzhangx: could refer to https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/2014365ea7b3405e024be1d9552aa30d44da4a8b/pkg/gce-pd-csi-driver/node.go#L108-L110

harshika-kashyap: sure.

harshika-kashyap:
@andyzhangx done with the changes

andyzhangx:
> @andyzhangx done with the changes

thanks, I would suggest use real lock, could refer to https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/2014365ea7b3405e024be1d9552aa30d44da4a8b/pkg/common/volume_lock.go#L44
though sync.Map is thread safe, while it cannot prevent race condition, use lock could solve the issue, thanks.

and it's better that `volumeLocks` could be in Driver struct fields, thanks.

pkg/azurefile/utils.go
```diff
@@ -43,15 +43,17 @@ func newLockMap() *lockMap {
...
-	}
+	defer lm.Unlock()
 
-	lm.Unlock()
+	// check if entry exists, then it is not possible to acquire lock
+	if _, exists := lm.mutexMap[entry]; exists {
+		return false
+	}
+	lm.addEntry(entry)
 	lm.lockEntry(entry)
```
harshika-kashyap, line 55: @andyzhangx Should we use `sets.String` instead of `mutexMap map[string]*sync.Mutex`? It will simplify the logic. 

Similar to: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/2014365ea7b3405e024be1d9552aa30d44da4a8b/pkg/common/volume_lock.go#L32

harshika-kashyap: Please ignore. I'm still updating this.

andyzhangx (commented):
/ok-to-test

andyzhangx (approved):
/lgtm
/approve
could you also port this feature to blob & smb csi driver? thanks.

PR #587, opened 2021-03-10
Bugfix for volume being forced to one account

edwardsp:
**What type of PR is this?**
/kind bug

A PV that fails to be created when a storage account reaches capacity will not get provisioned until space is freed.  The volume name needs to be removed from the volMap cache so that it can be placed in a different storage account.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


edwardsp:
CLA signed

PR #697, opened 2021-06-11
feat: Add networkEndpointType parameter to allow customers to create Private Endpoints

### LINKED ISSUE #633: support create storage account with private endpoint by the driver

support create storage account with private endpoint by the driver
Current behavior:
user should create storage account with private end point, and then specify storage account and private end point in storage class, here is an example: 
```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: azurefile-csi
parameters:
  resourceGroup: EXISTING_RESOURCE_GROUP_NAME  # optional, only set this when storage account is not in the same resource group as agent node
  storageAccount: EXISTING_STORAGE_ACCOUNT_NAME
  server: xxx  # e.g. accountname.privatelink.file.core.windows.net
```

**Additional context**

nearora-msft:
**What type of PR is this?**

/kind feature

This PR will allow customers to use networkEndpointType parameter to create private endpoints for storage account for NFS file shares.

Fixes #633 

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
feat: Add networkEndpointType parameter to allow customers to create Private Endpoints
```


nearora-msft:
@edreed for review

pkg/azurefile/controllerserver.go
```diff
@@ -152,6 +153,8 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 			}
 		case storageEndpointSuffixField:
 			storageEndpointSuffix = v
+		case networkEndpointTypeField:
+			networkEndpointType = v
```
andyzhangx, line 157: `createPrivateEndpoint = strings.EqualsFold(privateEndPointType, v)`

docs/driver-parameters.md
```diff
@@ -8,8 +8,9 @@ Name | Meaning | Example | Mandatory | Default value
...
+enableLargeFileShares | specify whether to use a storage account with large file shares enabled or not. If this flag is set to true and a storage account with large file shares enabled doesn't exist, a new storage account with large file shares enabled will be created. This flag should be used with the standard sku as the storage accounts created with premium sku have largeFileShares option enabled by default.  | `true`,`false` | No | `false`
 protocol | specify file share protocol | `smb`, `nfs` (`nfs` is in [Preview](https://github.com/kubernetes-sigs/azurefile-csi-driver/tree/master/deploy/example/nfs)) | No | `smb` 
+networkEndpointType | specify the network endpoint type to be created for the storage account. This parameter should only be used in the dynamic provisioning scenarios when a new storage account has to be created and the protocol is nfs. If the value of this parameter is 'privateendpoint', a private endpoint will be created for the storage account. For any other values or empty value of this parameter, a service endpoint will be created by default if the protocol is nfs. | "","privateendpoint" | No | ""
```
andyzhangx, line 13: `privateEndpoint`

andyzhangx (commented):
would you add an e2e test? thanks.

pkg/azurefile/controllerserver.go
nearora-msft: Should we keep it as it is? This will be more extensible in case there's another type of endpoint in future.

andyzhangx: make sense.
While for parameter value check, should always use `strings.EqualsFold` to make case insensitive.

test/e2e/dynamic_provisioning_test.go
```diff
@@ -142,6 +142,48 @@ var _ = ginkgo.Describe("Dynamic Provisioning", func() {
...
+	ginkgo.It("should create a volume on demand on a storage account with private endpoints  [kubernetes.io/azure-file] [file.csi.azure.com] [Windows]", func() {
+		skipIfUsingInTreeVolumePlugin()
+		pods := []testsuites.PodDetails{
+			{
+				Cmd: convertToPowershellCommandIfNecessary("echo 'hello world' > /mnt/test-1/data && grep 'hello world' /mnt/test-1/data"),
+				Volumes: []testsuites.VolumeDetails{
+					{
+						ClaimSize: "10Gi",
+						MountOptions: []string{
+							"dir_mode=0777",
```
andyzhangx, line 154: use nfs mount options instead, that's the reason why e2e test failed.

pkg/azurefile/controllerserver.go
nearora-msft: Changed to strings.EqualsFold

andyzhangx (approved):
/lgtm
/approve

PR #755, opened 2021-08-10
feat: Implement Volume_mount_group capability

### LINKED ISSUE #708: Implement gid support as per CSI 1.5
We should implement gid support as CSI 1.5 - https://github.com/container-storage-interface/spec/pull/468

xref https://github.com/kubernetes/enhancements/issues/2317

nearora-msft:
**What type of PR is this?**
/kind feature

Fixes #708 

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


nearora-msft:
@andyzhangx The sanity tests are failing because the newly added capability is still alpha and hasn't been added to the csi-sanity repo. Should we disable the test for now?

pkg/azurefile/nodeserver.go
```diff
@@ -142,6 +146,19 @@ func (d *Driver) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRe
...
 	context := req.GetVolumeContext()
 	mountFlags := req.GetVolumeCapability().GetMount().GetMountFlags()
+	volumeMountGroup := req.GetVolumeCapability().GetMount().GetVolumeMountGroup()
+	gidExistsInMountFlags := false
+	if volumeMountGroup != "" {
+		for _, mountFlag := range mountFlags {
+			if strings.HasPrefix(mountFlag, "gid") {
+				gidExistsInMountFlags = true
+				kvpair := strings.Split(mountFlag, "=")
+				if !strings.EqualFold(volumeMountGroup, kvpair[1]) {
```
andyzhangx, line 156: I would suggest add this logic in a separate function, that would be easier to write ut.

test/sanity/run-test.sh
```diff
@@ -48,4 +48,4 @@ fstype: ext4
 EOF
 
 echo 'Begin to run sanity test for vhd disk feature...'
-"$CSI_SANITY_BIN" --ginkgo.v --ginkgo.noColor --csi.endpoint="$endpoint" --csi.testvolumeparameters="$testvolumeparameters" --ginkgo.skip='should fail when the volume source snapshot is not found|should work|should fail when volume does not exist on the specified path|should fail when the volume does not exist|should fail when the node does not exist|should be idempotent'
+"$CSI_SANITY_BIN" --ginkgo.v --ginkgo.noColor --csi.endpoint="$endpoint" --csi.testvolumeparameters="$testvolumeparameters" --ginkgo.skip='should fail when the volume source snapshot is not found|should work|should fail when volume does not exist on the specified path|should fail when the volume does not exist|should fail when the node does not exist|should be idempotent|should return appropriate capabilities'
```
andyzhangx, line 51: where is the error `should return appropriate capabilities`? may fix it?

nearora-msft: The error is coming because we are implementing a new capability(Volume_Mount_Group) and this capability isn't recognized by the sanity tests(maybe because it's still an alpha feature). I am not sure if it makes sense to add this new capability to the sanity test considering that it is still an alpha feature. WTYD?

pkg/azurefile/nodeserver.go
nearora-msft: Done

pkg/azurefile/nodeserver.go
```diff
@@ -444,3 +456,19 @@ func makeDir(pathname string) error {
...
 }
+
+func checkGidPresentInMountFlags(volumeMountGroup string, mountFlags []string) (bool, error) {
+	gidPresentInMountFlags := false
+	if volumeMountGroup != "" {
+		for _, mountFlag := range mountFlags {
+			if strings.HasPrefix(mountFlag, "gid") {
+				gidPresentInMountFlags = true
+				kvpair := strings.Split(mountFlag, "=")
+				if !strings.EqualFold(volumeMountGroup, kvpair[1]) {
```
andyzhangx, line 467: there could be panic, use: 
`if len(kvpair) == 2 && !strings.EqualFold(volumeMountGroup, kvpair[1]) { }`

andyzhangx: write a ut for this function

test/sanity/run-test.sh
andyzhangx: only should skip `Node Service NodeGetCapabilities` now in both Line43 & Line51

pkg/azurefile/nodeserver.go
nearora-msft: Done

test/sanity/run-test.sh
nearora-msft: Done

andyzhangx (approved):
/lgtm
/approve

PR #874, opened 2022-01-06
test: change helm install method

### LINKED ISSUE #873: Improve download of dependencies

The script https://github.com/kubernetes-sigs/azurefile-csi-driver/blob/master/hack/verify-helm-chart-files.sh#L52 is doing a "curl <script> | bash" [anti-pattern](https://sysdig.com/blog/friends-dont-let-friends-curl-bash/).

It would be better supply chain security if:
- The script being downloaded was checked into the repo directly so that we can validate it is safe.
- The tool needed is installed through a package manager and/or have the checksums validated.

**Additional context**

andyzhangx:
**What type of PR is this?**
/kind test

test: change helm install method, we don't want to use `curl` even in testing which is "insecure"

Fixes #873

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


msau42:
Thank you!

PR #1013, opened 2022-05-07
feat: add fsGroupChangePolicy for nfs protocol

### LINKED ISSUE #1012: VHD disks don't respect fsgroup
I was testing the VHD feature of azurefile and it integrates kinda poorly with fsgroup specified in pod. Currently it looks like, if volume is in diskMode, it ignores supplied `fsGroup` completely - https://github.com/kubernetes-sigs/azurefile-csi-driver/blob/master/pkg/azurefile/nodeserver.go#L248

But even if it did, since now we are mounting the cifs file and then creating a loopback device on it, ideally it is kubelet which must recursively chown and chmod files. 

This is basically a side effect of exposing multiple volume types under same CSI driver name and will be tricky to fix.

andyzhangx:
**What type of PR is this?**
/kind bug
/kind feature

fix: fsgroup setting on vhd disk
feat: add fsGroupChangePolicy for nfs protocol

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
fix: fsgroup setting on vhd disk
feat: add fsGroupChangePolicy for nfs protocol
```

### DIFF
--- a/pkg/azurefile/nodeserver.go
+++ b/pkg/azurefile/nodeserver.go
@@ -167,7 +167,7 @@ func (d *Driver) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRe
 	}
 	// don't respect fsType from req.GetVolumeCapability().GetMount().GetFsType()
 	// since it's ext4 by default on Linux
-	var fsType, server, protocol, ephemeralVolMountOptions, storageEndpointSuffix, folderName string
+	var fsType, server, protocol, ephemeralVolMountOptions, storageEndpointSuffix, folderName, fsGroupChangePolicy string
 	var ephemeralVol bool
 	mountPermissions := d.mountPermissions
 	performChmodOp := (mountPermissions > 0)
@@ -189,6 +189,8 @@ func (d *Driver) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRe
 			ephemeralVolMountOptions = v
 		case storageEndpointSuffixField:
 			storageEndpointSuffix = v
+		case fsGroupChangePolicyField:
+			fsGroupChangePolicy = v
 		case mountPermissionsField:
 			if v != "" {
 				var err error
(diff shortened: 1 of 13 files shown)


pkg/azurefile/nodeserver.go
```diff
@@ -329,6 +343,15 @@ func (d *Driver) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRe
 		}
 		klog.V(2).Infof("NodeStageVolume: volume %s format %s and mounting at %s successfully", volumeID, targetPath, diskPath)
 	}
+
+	if protocol == nfs || isDiskMount {
+		if volumeMountGroup != "" {
+			klog.V(2).Infof("set gid of volume(%s) as %s using fsGroupChangePolicy(%s)", volumeID, volumeMountGroup, fsGroupChangePolicy)
```
gnufied, line 349: @andyzhangx did you see my comment https://github.com/kubernetes-sigs/azurefile-csi-driver/issues/1012#issuecomment-1120196467 ? At least to me this does not seem right and we merged it.

andyzhangx: hi @gnufied, this is a workaround fix since nfs and smb are both supported with this driver. Is there any better way to fix this?

gnufied: There is no better way to fix this for VHD disks, but I am really concerned that if we take path taken in this code then customers will want to rely on it and when with a future version of CSI and k8s these changes are finalized - it will break their work-flow.

andyzhangx: thanks for raising the concern. currently `fsGroupChangePolicy` is an azure file volume property, I think we could do some backwards compatibility with future version of CSI and k8s these changes are finalized, we already hit some mount issues recently and need to workaround ASAP. thanks.

Name | Meaning | Example | Mandatory | Default value 
--- | --- | --- | --- | ---
fsGroupChangePolicy | indicates how volume's ownership will be changed by the driver, pod `securityContext.fsGroupChangePolicy` is ignored  | `OnRootMismatch`(by default), `Always`, `None` | No | `OnRootMismatch`

PR #1023, opened 2022-05-18
add new option to allow VHD feature opt-out

### LINKED ISSUE #1012: VHD disks don't respect fsgroup

RomanBednar:
**What type of PR is this?**
/kind bug
/kind regression

Currently when users configure Azure file storage class with fstype param any non-admin users will get permission denied errors.

```
kind: StorageClass
apiVersion: storage.k8s.io/v1
metadata:
  name: mystorageclass
parameters:
  fstype: ext4
reclaimPolicy: Delete
allowVolumeExpansion: true
volumeBindingMode: WaitForFirstConsumer
```

When trying to access the volume on a pod:
```
Can not read/write into mounted volume due to "permission denied Error"
 ERROR: cannot create regular file '/mnt/storage/hello': Permission denied
```

Fixes #1012 

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
users can now opt-out of VHD disk feature using new option --enable-vhd (default: true)
```

### DIFF
--- a/pkg/azurefileplugin/main.go
+++ b/pkg/azurefileplugin/main.go
@@ -51,6 +51,7 @@ var (
 	mountPermissions                       = flag.Uint64("mount-permissions", 0777, "mounted folder permissions")
 	allowInlineVolumeKeyAccessWithIdentity = flag.Bool("allow-inline-volume-key-access-with-identity", false, "allow accessing storage account key using cluster identity for inline volume")
 	fsGroupChangePolicy                    = flag.String("fsgroup-change-policy", "", "indicates how the volume's ownership will be changed by the driver, OnRootMismatch is the default value")
+	enableVHDDiskFeature                   = flag.Bool("enable-vhd", true, "enable VHD disk feature (experimental)")
 )
 
 func main() {
@@ -87,6 +88,7 @@ func handle() {
 		MountPermissions:                       *mountPermissions,
 		AllowInlineVolumeKeyAccessWithIdentity: *allowInlineVolumeKeyAccessWithIdentity,
 		FSGroupChangePolicy:                    *fsGroupChangePolicy,
+		EnableVHDDiskFeature:                   *enableVHDDiskFeature,
 	}
 	driver := azurefile.NewDriver(&driverOptions)
 	if driver == nil {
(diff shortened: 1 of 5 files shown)


andyzhangx:
thanks for the contribution.
I think the permission issue should be fixed by this PR: https://github.com/kubernetes-sigs/azurefile-csi-driver/pull/1013

if fsType is ext4, then it would set group id on the volume:
https://github.com/kubernetes-sigs/azurefile-csi-driver/blob/9e49a8e7dd55ac92101685895d3484a36659ecc5/pkg/azurefile/nodeserver.go#L347-L353

RomanBednar:
> thanks for the contribution. I think the permission issue should be fixed by this PR: #1013

I see, I was not aware of this PR, thank you.

jsafrane:
I think this PR and #1013 are orthogonal. If block volumes are still preview, it should be possible to turn it off (or on).

/reopen

pkg/azurefileplugin/main.go
```diff
@@ -51,6 +51,7 @@ var (
 	mountPermissions                       = flag.Uint64("mount-permissions", 0777, "mounted folder permissions")
 	allowInlineVolumeKeyAccessWithIdentity = flag.Bool("allow-inline-volume-key-access-with-identity", false, "allow accessing storage account key using cluster identity for inline volume")
 	fsGroupChangePolicy                    = flag.String("fsgroup-change-policy", "", "indicates how the volume's ownership will be changed by the driver, OnRootMismatch is the default value")
+	experimentalEnableFsType               = flag.Bool("experimental-enable-fstype", false, "enable fstype storage class parameter usage (experimental)")
```
andyzhangx, line 54: we should set as `true` by default

RomanBednar: Done. I've also renamed the option so it better describes what it actually does.

andyzhangx (approved):
/lgtm
/approve

PR #1030, opened 2022-05-25
doc: mark VHD disk feature as experimental feature

andyzhangx:
**What type of PR is this?**
/kind documentation

doc: mark VHD disk feature as experimental feature

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

/assign @jsafrane


docs/driver-parameters.md
```diff
@@ -37,7 +37,7 @@ vnetResourceGroup | specify vnet resource group where virtual network is | exist
 vnetName | virtual network name | existing virtual network name | No | if empty, driver will use the `vnetName` value in azure cloud config file
 subnetName | subnet name | existing subnet name of the agent node | No | if empty, driver will use the `subnetName` value in azure cloud config file
 fsGroupChangePolicy | indicates how volume's ownership will be changed by the driver, pod `securityContext.fsGroupChangePolicy` is ignored  | `OnRootMismatch`(by default), `Always`, `None` | No | `OnRootMismatch`
---- | **Following parameters are only for [VHD disk feature](../deploy/example/disk)** | --- | --- |
```
gnufied, line 40: I was thinking having a feature-gate that needs to be required to enabled to use the feature. Something akin to how main k8s/k8s experimental/alpha features work.

gnufied: If the feature gate marks it as beta and is enabled by default, that works too. This way, if any other k8s distro wants to disable the feature they can do so.

Issue #1062, opened 2022-07-28
Reference image digests instead of tags during install


grtn316:
**Is your feature request related to a problem?/Why is this needed**
Digests are immutable while tags can be repointed at different digests. Using tags could also present security risks or operational impacts because it is not obvious to the user that actual image has changed but the tag has not.

Openshift 4+ does not support tag references when mirroring public repos to customers private repos (for clusters without internet access). Digest references are required instead. You can implement a work around, however, it is not recommended by RedHat and is a bit cumbersome to implement.

**Describe the solution you'd like in detail**
When installing the CSI drivers, reference the digest instead of the tag. For example:
```bash
podman pull quay.io/mongodb/mongodb-kubernetes-operator@sha256:e19ae43539521f0350fb71684757dc535fc989deb75f3789cd84b782489eda80
```

Issue #1020, opened 2022-05-17
Deploying to ARO - azure-cloud-provider encoding


grtn316:
When deploying to ARO, there is already an azure-cloud-provider secret. This secret is plain JSON and not base64'd like it says to do in the docs when creating a new one:
```bash
export AZURE_CLOUD_SECRET=cat deploy/cloud.conf | base64 | awk '{printf $0}'; echo
```

I was updating the file to include:
```json 
{
    "cloud":"AzurePublicCloud",
    "tenantId": "$tenantId",
    "subscriptionId": "$subscriptionId",
    "resourceGroup": "$resourceGroupName",
    "location": "$deployRegion",
    "aadClientId": "$clientId",
    "aadClientSecret": "$clientSecret",
    "useManagedIdentityExtension": false,
    "userAssignedIdentityID": "",
    "useInstanceMetadata": true,
    "vmType": "standard",
    "subnetName": "$subnetPrivateEndpointName",
    "vnetName": "$vnetName",
    "vnetResourceGroup": "$resourceGroupName",
    "cloudProviderBackoff": true
}
```

I initially stored the base64 version which kept failing. After adding the plain JSON object into the secret, the drivers began to work. I did not notice this documented in the ARO docs anywhere so I figured I would mention it.

I am using driver version **v1.12.0**

andyzhangx:
do you have the driver controller logs to share, follow https://github.com/kubernetes-sigs/azurefile-csi-driver/blob/master/docs/csi-debug.md#case1-volume-createdelete-issue

grtn316:
Turns out this is not specific to ARO. I have also discovered it on older version of openshift (4.8.42) and thinking its related to the image versions instead. For example: I just deployed 1.12.0 and am seeing this in the driver controller logs:

```
I0728 18:22:48.683765       1 azurefile.go:259] driver userAgent: file.csi.azure.com/v1.12.0 gc/go1.17 (amd64-linux) OSS-kubectl
I0728 18:22:48.684216       1 azure.go:71] reading cloud config from secret kube-system/azure-cloud-provider
W0728 18:22:48.696911       1 azure.go:78] InitializeCloudFromSecret: failed to get cloud config from secret kube-system/azure-cloud-provider: failed to parse Azure cloud-config: error unmarshaling JSON: while decoding JSON: json: cannot unmarshal string into Go value of type provider.Config
[...]

PR #1063, opened 2022-07-28
For v1.12.0 pull images by digest instead of tags

grtn316:
**What type of PR is this?**
/kind feature
/kind bug

feature: version 1.12.0 supports pulling images using digests instead of tags. This prevents pointing the tag to another digest and causing an accidental update in production scenarios.

bug: The installer is pointed at the version branch instead of the master branch. I have updated it to point to /master/depoy/$ver instead of /$ver/deploy/$ver. This makes it possible for forks to use their localized versions correctly without having to fork all of the version branches as well.

Fixes #

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


andyzhangx:
hi @grtn316 thanks for the contribution. the image under `mcr.microsoft.com/k8s/csi/azurefile-csi/` is immutable after publish, so it's not necessary to use digest, also using tags could easily show which version it is, the k8s upstream are also using tags in csi driver deployment.

grtn316:
Hi @andyzhangx while I may not have all the details you require to magnify why using digests for the actual install is important, I will attempt to spotlight some examples and current issues that may help:

The images currently in the repo: `mcr.microsoft.com/k8s/csi/azurefile-csi/` are tagged but also have digests. You can either reference the tag or the digest in the deployment yamls.

**Risks of using tags in the deployment:**

- Underlying image for the tag can technically be changed without the cluster administrator knowing. This control is in the hands of the image owner versus the cluster owner (since they are using the install scripts that reference the tag). Resulting in randomly failed pods in production when the cluster admin is doing maintenace. I believe this could be the cause of this issue: #1020. We have not changed anything on our side and the issue suddenly showed up. We have been pinned to v1.12.0 since it came out. I may be able to recover an old cluster to look at the image:tag on one of the nodes to see if the digest matches the current image that is tagged in `mcr.microsoft.com/k8s/csi/azurefile-csi/` but I suspect the underlying image for tag: 1.12.0 may have gotten updated at some point to fix a bug or security vulnerability and accidently included the change discovered in the #1020 issue.
- When using tags, mirrors or malicious mirrors are able to republish a malicious image with the same name:tag and cluster admins could accidently pull down this image causing a security vulnerability.

**Benefits of using digest references in the deployment yamls:**

- When the underlying image for a tag is updated/changed the cluster will not pull the new version as it is pinned to an immutable digest for that specific build.
- Malicious images would have a new digest reducing the immediately blast radius since clusters do not target the new digest.

**Comments:**
I was able to mirror the images from `mcr.microsoft.com/k8s/csi/azurefile-csi/` using skopeo to an azure container registry and update the deployment yamls to target the digests for v1.12.0 instead of the tags (as seen in this PR). Tags can still be used for convenience but referencing the digest for the deployment creates a more consistent and secure deployment for users.

If my suspicion is correct for #1020 then this would invalidate that once an image is tagged in `mcr.microsoft.com/k8s/csi/azurefile-csi/` that it will never be updated. Since the tag pointer is not technically immutable, I suspect you are referring to the process of how tags are currently maintained in that repository.

I also created a feature request here to address the digest topic #1062

PR #1077, opened 2022-08-31
feat: support cross subscription with management api

### LINKED ISSUE #1066: SubscriptionID parameter not working
Version: 1.20.0

Hi we have an AKS Cluster deployed in Subscription "B" with Resource Group "B_AKS"
We want to Mount azurefiles from Subscription "A" with Resource Group "A_Base"
The storageAccount is named "BaseStorage"
The AKS-SP-Identity is assigned as Contributer to ResourceGroup "A_Base" in subscription "A"

When setting the storage-class paramters like:
```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: baserg-azurefile-csi  
allowVolumeExpansion: true
parameters:
  subscriptionID: A
  resourceGroup: A_Base
  storageAccount: BaseStorage
reclaimPolicy: Delete
volumeBindingMode: Immediate
mountOptions:
  - dir_mode=0777
  - file_mode=0777
  - uid=0
  - gid=0
  - mfsymlinks
  - cache=strict  # https://linux.die.net/man/8/mount.cifs
  - nosharesock  # reduce probability of reconnect race
  - actimeo=30  # reduce latency for metadata-heavy workload
```

The provisioning fails with a message saying unauthorized or invalid scope:
...does not have authorization to perform action 'Microsoft.Storage/storageAccounts/fileServices/shares/read' over scope '/subscriptions/B/resourceGroups/A_Base/providers/Microsoft.Storage/storageAccounts/BaseStorage/fileServices/default/shares/pvc-b5511b84-9f65-451f-9a88-43787dcdc478' or the scope is invalid

As you can see the requested subscription is ***B*** and not A
Is this a bug or are we doing something wrong?

cvvz:
**What type of PR is this?**
/kind feature

support create azure file across subscription without using data plane api

Fixes #1066

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
support create azure file across subscription without using data plane api

```


cvvz:
@andyzhangx  @ZeroMagic  PTAL while I'm still testing for the functionality locally.

andyzhangx (commented):
/ok-to-test

pkg/azurefile/azurefile.go
```diff
@@ -371,27 +371,36 @@ func (d *Driver) getFileShareQuota(resourceGroupName, accountName, fileShareName
...
 		// in csi migration, rg could be empty, then the 5th element is namespace
 		// https://github.com/kubernetes/kubernetes/blob/v1.23.5/staging/src/k8s.io/csi-translation-lib/plugins/azure_file.go#L137
-		namespace = segments[4]
-	}
-	if len(segments) > 5 {
-		namespace = segments[5]
+		if len(segments) > 4 {
+			namespace = segments[4]
+		}
+		if len(segments) > 5 {
```
andyzhangx, line 392: should be like this:
```
	if rg == "" {
		// in csi migration, rg could be empty, then the 5th element is namespace
		// https://github.com/kubernetes/kubernetes/blob/v1.23.5/staging/src/k8s.io/csi-translation-lib/plugins/azure_file.go#L137
		if len(segments) > 4 {
			namespace = segments[4]
		}
	} else {
		if len(segments) > 5 {
			namespace = segments[5]
		}
	}
	if len(segments) > 6 {
		subsID = segments[6]
	}
```

andyzhangx: also write a ut for `rg == ""` condition

andyzhangx (commented):
pls also squash all commits, thanks.

pkg/azurefile/azurefile.go
andyzhangx: csi migration does not support cross subs, we don't need to consider such scenario

cvvz: fixed it, and there is already an ut covered `rg == ""` condition

pkg/azurefile/controllerserver.go
```diff
@@ -797,7 +807,18 @@ func (d *Driver) CreateSnapshot(ctx context.Context, req *csi.CreateSnapshotRequ
 		return nil, status.Error(codes.InvalidArgument, "CreateSnapshot Source Volume ID must be provided")
 	}
 
-	mc := metrics.NewMetricContext(azureFileCSIDriverName, "controller_create_snapshot", d.cloud.ResourceGroup, d.cloud.SubscriptionID, d.Name)
+	rgName, _, _, _, _, subsID, err := GetFileShareInfo(sourceVolumeID)
```
andyzhangx, line 810: maybe it's better to return a struct pointer here since there is already golint error here: `declaration has 4 blank identifiers`

cvvz: Could I workaround it by adding comment to exclude these `dogsled` linters this time? Since too many places to fix which raises the risk of making mistake and makes the code review more difficult, while the improvement to the code is very limited.

pkg/azurefile/controllerserver.go
```diff
@@ -218,12 +218,16 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		return nil, status.Errorf(codes.InvalidArgument, fmt.Sprintf("matchTags must set as false when storageAccount(%s) is provided", account))
 	}
 
-	if subsID != "" && subsID != d.cloud.SubscriptionID {
+	if subsID == "" {
+		subsID = d.cloud.SubscriptionID
+	}
+
+	if subsID != d.cloud.SubscriptionID {
 		if fsType == nfs || protocol == nfs {
```
andyzhangx, line 226: I think we could remove this restriction also since  nfs should support cross subs.

pkg/azurefile/controllerserver.go
```diff
@@ -529,7 +533,7 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		// not necessary for dynamic file share name creation since volumeID already contains volume name
 		uuid = volName
 	}
-	volumeID = fmt.Sprintf(volumeIDTemplate, resourceGroup, accountName, validFileShareName, diskName, uuid, secretNamespace)
+	volumeID = fmt.Sprintf(volumeIDTemplate, resourceGroup, accountName, validFileShareName, diskName, uuid, secretNamespace, subsID)
```
andyzhangx, line 536: current volumeid contains subsid, is it possible to not append subsid by default, we only append subsid when it's cross subs?

`volumeid="kubetest-uyg5f4ur#fb2e232232e1d4e799ac350#pvc-ac1d3f51-9ed4-43c6-9ef0-486d09f05e3f###kube-system#0e46bd28-a80f-4d3a-8200-d9eb8d80cb2e"`

andyzhangx: What about this? keep `volumeIDTemplate` unchanged, and then
```
if subsID != "" {
  volumeID = volumeID + "#" + subsID
} 
```

by this way, volumeID format is not changed if it's not cross subs scenario.

cvvz: That's right

andyzhangx (approved):
/approve
could you squash all commits? thanks.

ZeroMagic (approved):
/lgtm

PR #1085, opened 2022-09-13
fix: cancel the context when timeout of deleting file share

### LINKED ISSUE #1083: Failed to delete file share when timeout
**What happened**:
When the response error from [DeleteFileShare](https://github.com/kubernetes-sigs/azurefile-csi-driver/blob/master/pkg/azurefile/controllerserver.go#L600) is `volume deletion failed: rpc error: code = DeadlineExceeded desc = context deadline exceeded`, the goroutine would die, and the volume would be [locked](https://github.com/kubernetes-sigs/azurefile-csi-driver/blob/master/pkg/azurefile/controllerserver.go#L563-L566) in driver. 
Then the next time deletion retry by external-provisioner, we would get the following error.
```
volume deletion failed: rpc error: code = Aborted desc = An operation with the given Volume ID xxx already exists
```

Successfully retry the deletion.

**How to reproduce it**:

The essential issue is that the context wasn't cancelled when timeout. So we can reuse the context from external-provisioner. The field "timeout" in this context is filled.

https://github.com/kubernetes-csi/external-provisioner/blob/b3d0a89b38dede277ea1a5c200f3da6a328bc3e7/pkg/controller/controller.go#L1840-L1852

```
deleteCtx, cancel := context.WithTimeout(ctx, p.timeout)
	defer cancel()
	for i := 0; i < deleteVolumeRetryCount; i++ {
		_, err = p.csiClient.DeleteVolume(deleteCtx, delReq)
		if err == nil {
			break
		}
	}
```

**Environment**:

- CSI Driver version:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

ZeroMagic:
**What type of PR is this?**

/kind bug

As the description in the issue https://github.com/kubernetes-sigs/azurefile-csi-driver/issues/1083, here we reuse the context from external-provisioner to cancel the context when timeout. Then the volume won't be locked in driver.

Fixes #1083 

**Requirements**:
  
- [x] includes documentation
- [x] adds unit tests

**Release note**:
```
fix: cancel the context when timeout of deleting file share
```

### DIFF
--- a/pkg/azurefile/azurefile.go
+++ b/pkg/azurefile/azurefile.go
@@ -755,7 +755,7 @@ func (d *Driver) CreateFileShare(accountOptions *azure.AccountOptions, shareOpti
 }
 
 // DeleteFileShare deletes a file share using storage account name and key
-func (d *Driver) DeleteFileShare(subsID, resourceGroup, accountName, shareName string, secrets map[string]string) error {
+func (d *Driver) DeleteFileShare(ctx context.Context, subsID, resourceGroup, accountName, shareName string, secrets map[string]string) error {
 	return wait.ExponentialBackoff(d.cloud.RequestBackoff(), func() (bool, error) {
 		var err error
 		if len(secrets) > 0 {
@@ -763,7 +763,7 @@ func (d *Driver) DeleteFileShare(subsID, resourceGroup, accountName, shareName s
 			if rerr != nil {
 				return true, rerr
 			}
-			err = d.fileClient.deleteFileShare(accountName, accountKey, shareName)
+			err = d.fileClient.deleteFileShare(ctx, accountName, accountKey, shareName)
 		} else {
 			err = d.cloud.DeleteFileShare(subsID, resourceGroup, accountName, shareName)
 		}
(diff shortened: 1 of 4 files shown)


pkg/azurefile/azurefile.go
```diff
@@ -755,15 +755,15 @@ func (d *Driver) CreateFileShare(accountOptions *azure.AccountOptions, shareOpti
...
+func (d *Driver) DeleteFileShare(ctx context.Context, subsID, resourceGroup, accountName, shareName string, secrets map[string]string) error {
 	return wait.ExponentialBackoff(d.cloud.RequestBackoff(), func() (bool, error) {
 		var err error
 		if len(secrets) > 0 {
 			accountName, accountKey, rerr := getStorageAccount(secrets)
 			if rerr != nil {
 				return true, rerr
 			}
-			err = d.fileClient.deleteFileShare(accountName, accountKey, shareName)
+			err = d.fileClient.deleteFileShare(ctx, accountName, accountKey, shareName)
```
andyzhangx, line 766: I think original issue is the test is running on L768: `d.cloud.DeleteFileShare(subsID, resourceGroup, accountName, shareName)` it's stuck there.

andyzhangx: thanks, let's merge this PR first.

andyzhangx (approved):
/lgtm
/approve

PR #1097, opened 2022-09-26
[DO NOT MERGE] remove csi proxy

ZeroMagic:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/mounter/safe_mounter_windows.go
```diff
@@ -88,17 +59,64 @@ func (mounter *csiProxyMounter) SMBMount(source, target, fsType string, mountOpt
...
-		RemotePath: source,
-		Username:   mountOptions[0],
-		Password:   sensitiveMountOptions[0],
-	}
+
 	klog.V(2).Infof("begin to mount %s on %s", source, normalizedTarget)
-	if _, err := mounter.SMBClient.NewSmbGlobalMapping(context.Background(), smbMountRequest); err != nil {
-		return fmt.Errorf("smb mapping failed with error: %v", err)
+
+	klog.V(2).Infof("calling NewSmbGlobalMapping with remote path %q", source)
```
andyzhangx, line 65: for POC, it's ok, in long term, should implement a new interface for `SMBMount` native call

deploy/csi-azurefile-node-windows-hostprocess.yaml
```diff
@@ -0,0 +1,141 @@
...
+                  apiVersion: v1
+                  fieldPath: spec.nodeName
+          volumeMounts:
+            - name: pwsh
+              mountPath: "C:\\WindowsPowerShell"
+      volumes:
+        - name: pwsh
+          hostPath:
+            path: C:\Windows\System32\WindowsPowerShell\v1.0
+            type: DirectoryOrCreate
```
ZeroMagic, line 141: @mauriciopoppe @alexander-ding How do you handle the powershell command inside container in your library solution? 
After testing, the powershell command cannot be found inside container in spite of using powershell related image. So here I mount powershell from host.

andyzhangx: I think the point is it's always using `C:\Windows\System32\WindowsPowerShell\v1.0`, and `v1.0` may change in the future, is there any way to avoid this though it works now?

pkg/azurefileplugin/main.go
```diff
@@ -54,6 +54,7 @@ var (
 	enableVHDDiskFeature                   = flag.Bool("enable-vhd", true, "enable VHD disk feature (experimental)")
 	kubeAPIQPS                             = flag.Float64("kube-api-qps", 25.0, "QPS to use while communicating with the kubernetes apiserver.")
 	kubeAPIBurst                           = flag.Int("kube-api-burst", 50, "Burst to use while communicating with the kubernetes apiserver.")
+	enableWindowsHostProcess               = flag.Bool("enableWindowsHostProcess", false, "enable windows host process")
```
andyzhangx, line 57: `enable-windows-host-process`

Makefile
```diff
@@ -100,11 +100,13 @@ ifdef TEST_WINDOWS
 		--set driver.azureGoSDKLogLevel=INFO \
 		--set controller.replicas=1 \
 		--set controller.logLevel=6 \
-		--set node.logLevel=6
+		--set node.logLevel=6 \
+		--disable-openapi-validation
```
andyzhangx, line 104: there is `WINDOWS_USE_HOST_PROCESS_CONTAINERS` env var in test-infra, we could only `disable-openapi-validation` when `WINDOWS_USE_HOST_PROCESS_CONTAINERS` is set and also run host process test in e2e test:

https://github.com/kubernetes/test-infra/blob/b02ce7741ab34242f95f33fe680d8507c3060f09/config/jobs/kubernetes-sigs/azurefile-csi-driver/azurefile-csi-driver-config.yaml#L663-L664

pkg/azurefile/azurefile.go
```diff
@@ -177,6 +177,8 @@ var (
 	supportedFSGroupChangePolicyList = []string{FSGroupChangeNone, string(v1.FSGroupChangeAlways), string(v1.FSGroupChangeOnRootMismatch)}
 
 	retriableErrors = []string{accountNotProvisioned, tooManyRequests, shareBeingDeleted, clientThrottled}
+
+	enableWindowsHostProcess = false
```
andyzhangx, line 181: I think this var is useless

pkg/mounter/safe_mounter_host_process_windows.go
```diff
@@ -0,0 +1,265 @@
...
+
+	"sigs.k8s.io/azurefile-csi-driver/pkg/os/filesystem"
+	ossmb "sigs.k8s.io/azurefile-csi-driver/pkg/os/smb"
+)
+
+var _ CSIProxyMounter = &winMounter{}
+
+type winMounter struct{}
+
+func NewWinMounter() *winMounter {
```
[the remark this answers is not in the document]
andyzhangx: `NewWinNativeCallMounter `

deploy/csi-azurefile-node-windows-hostprocess.yaml
alexander-ding: @ZeroMagic there's no need to mount anything. HostProcess containers have no file system isolation and have a complete view of the host machine's file system, so Poweshell should always be available at a well known path. There are a few options to expose powershell, besides what you're doing:
 
1. Expose the powershell folder in PATH. This is a bit ugly, but it should make things work.
2. Build your base image from the specialized [HostProcess container base image](https://github.com/microsoft/windows-host-process-containers-base-image) provided by Microsoft. This image is much smaller and works for all Windows server versions. Importantly, it makes powershell available directly as a command.

I'd recommend (2). This is what we're doing in the GCP Compute CSI Driver too.

deploy/csi-azurefile-node-windows-hostprocess.yaml
```diff
@@ -0,0 +1,141 @@
...
+                    operator: NotIn
+                    values:
+                      - virtual-kubelet
+      priorityClassName: system-node-critical
+      securityContext:
+        windowsOptions:
+          hostProcess: true
+          runAsUserName: "NT AUTHORITY\\SYSTEM"
+      hostNetwork: true
+      initContainers:
```
alexander-ding, line 42: By the way, I don't think you need to use a full blown initContainer. You can just add a hostPath volume of type DirectoryOrCreate pointing to `/var/lib/kubelet/plugins/file.csi.azure.com`. This should prompt K8s to create the directory for you.

pkg/os/filesystem/types.go
```diff
@@ -0,0 +1,143 @@
...
+	// UNC paths of the form "\\server\share\path\file" are not allowed.
+	// All directory separators need to be backslash character: "\".
+	// Characters: .. / : | ? * in the path are not allowed.
+	// target_path needs to exist as a directory in the host that is empty.
+	// target_path cannot be a symbolic link.
+	// Maximum path length will be capped to 260 characters.
+	TargetPath string `protobuf:"bytes,2,opt,name=target_path,json=targetPath,proto3" json:"target_path,omitempty"`
+}
+
+type CreateSymlinkRequest struct {
```
andyzhangx, line 101: remove XXXRequest struct, and invoke directly in the safe_mounter_host_process.go would be simpler

deploy/csi-azurefile-node-windows-hostprocess.yaml
ZeroMagic: Will try it later.

PR #1145, opened 2022-12-23
docs: recommend the use of the nosharesock mount option, and clearify how volumeHandle should be used.

### LINKED ISSUE #1137: Stacked NodeStageVolume mounts prevents the kubelet from ever unmounting Azure Files volumes
**What happened**:
I have encountered a situation where our nodes are unable to unmount NodeStageVolume mounts of azure files shares. Once one of these mounts are made to the node, they stay there until the node is shutdown. This resulted in over 1400 mounts on several of our production nodes where we amongst other things had cron-jobs that mounted the same Azure Files share.

I eventually found out that we where not using the recommended`nosharesock` mount option - with that option added we did not get the problem and we're in the process of rolling this out.

But, I think that this should have worked even without `nosharesock`. It seems dangerous that it is ever possible for volumes to get stuck on a node like this. I have some more details on what we saw that can hopefully help to determine what actually is wrong.

When I logged on to a node where I had the problem reproduced I saw two "globalmounts" (aka NodeStageVolume mounts) for each of our Persistent Volume Claims associated to pods that has been previously scheduled to the node. 

Example data from `/proc/self/mountinfo` on a node that had two pods deployed that each mounted the same azure-files share once. (with some redaction)

```
1537 318 0:204 / /var/lib/kubelet/plugins/kubernetes.io/csi/file.csi.azure.com/6a0ab086793b5c17ad4f33b0441a5b0a5f008e150cd47e2cdc62f5dd8cd68cea/globalmount rw,relatime shared:368 - cifs //*****.file.core.windows.net/site-data rw,vers=3.1.1,cache=strict,username=*****,uid=33,forceuid,gid=33,forcegid,addr=52.239.143.8,file_mode=0777,dir_mode=0777,soft,persistenthandles,nounix,serverino,mapposix,nobrl,mfsymlinks,rsize=1048576,wsize=1048576,bsize=1048576,echo_interval=60,actimeo=33
1536 30 0:204 / /var/lib/kubelet/plugins/kubernetes.io/csi/file.csi.azure.com/6a0ab086793b5c17ad4f33b0441a5b0a5f008e150cd47e2cdc62f5dd8cd68cea/globalmount rw,relatime shared:368 - cifs //*****.file.core.windows.net/site-data rw,vers=3.1.1,cache=strict,username=*****,uid=33,forceuid,gid=33,forcegid,addr=52.239.143.8,file_mode=0777,dir_mode=0777,soft,persistenthandles,nounix,serverino,mapposix,nobrl,mfsymlinks,rsize=1048576,wsize=1048576,bsize=1048576,echo_interval=60,actimeo=33
1832 318 0:204 / /var/lib/kubelet/plugins/kubernetes.io/csi/file.csi.azure.com/2069f851ff9511cf4a09449f85f94ed1fe837836bd172bff2bbd0084eab267e4/globalmount rw,relatime shared:433 - cifs //*****.file.core.windows.net/site-data rw,vers=3.1.1,cache=strict,username=*****,uid=33,forceuid,gid=33,forcegid,addr=52.239.143.8,file_mode=0777,dir_mode=0777,soft,persistenthandles,nounix,serverino,mapposix,nobrl,mfsymlinks,rsize=1048576,wsize=1048576,bsize=1048576,echo_interval=60,actimeo=33
1831 30 0:204 / /var/lib/kubelet/plugins/kubernetes.io/csi/file.csi.azure.com/2069f851ff9511cf4a09449f85f94ed1fe837836bd172bff2bbd0084eab267e4/globalmount rw,relatime shared:433 - cifs //*****.file.core.windows.net/site-data rw,vers=3.1.1,cache=strict,username=*****,uid=33,forceuid,gid=33,forcegid,addr=52.239.143.8,file_mode=0777,dir_mode=0777,soft,persistenthandles,nounix,serverino,mapposix,nobrl,mfsymlinks,rsize=1048576,wsize=1048576,bsize=1048576,echo_interval=60,actimeo=33
```
In other words, two "globalmounts" per mounted volume. One seems to be [a parent of the other](https://github.com/kubernetes-sigs/azurefile-csi-driver/blob/4f66493c76b05e1efedfa59fb976b7ce9849d825/vendor/k8s.io/mount-utils/mount_helper_unix.go#L69)

When looking at the kubelet log it is filled with the following type of messages (from another debugging sessions so the id's wont match, also I've redacted some things with *****)
```
E1210 14:51:59.241317    1362 nestedpendingoperations.go:335] Operation for "{volumeName:kubernetes.io/csi/file.csi.azure.com^***** 
podName: nodeName:}" failed. No retries permitted until 2022-12-10 14:51:59.741298734 +0000 UTC m=+2014125.708638693 
(durationBeforeRetry 500ms). Error: GetDeviceMountRefs check failed for volume "*****" (UniqueName: 
"kubernetes.io/csi/file.csi.azure.com^*****") on node "aks-*****-24994263-vmss000000" : the device mount path 
"/var/lib/kubelet/plugins/kubernetes.io/csi/pv/*****/globalmount" is still mounted by other references 
[/var/lib/kubelet/plugins/kubernetes.io/csi/pv/*****/globalmount /var/lib/kubelet/plugins/kubernetes.io/csi/pv/*****/globalmount 
/var/lib/kubelet/plugins/kubernetes.io/csi/pv/*****/globalmount]
```
As far as I can tell the double mounts [trips up the kubelet duing cleanup](https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/util/operationexecutor/operation_generator.go#L965) as none of the functions that lists volumes seems to handler parent/child moutns. The result is that volume never gets unmounted.

The Kubelet + csi-driver should have been able to unmount the volumes as the pods left the node. Alternatively if mounts without `nosharesock` is not supported the csi-driver should have refused to mount it.

I could not find the place where the two layers of globalmounts gets created, so I could not say for sure whether it is by design or not. But if it is, we seem to be lacking some code to handle the deduplicated in the lists the kubelet uses to determine if a volume is still in use.

**How to reproduce it**:

1. Follow this guide for settings up a static Azure Files volume: https://learn.microsoft.com/en-us/azure/aks/azure-files-volume make sure to use the [Mount file share as a persistent volume](https://learn.microsoft.com/en-us/azure/aks/azure-files-volume#mount-file-share-as-a-persistent-volume) subsection but don't specify the `nosharesock` mount option from the Persistent Volume.

2. Open a shell into the node (i use a [krew plugin](https://github.com/kvaps/kubectl-node-shell) to do this) and list the mounts with eg. `mount` or `cat /proc/self/mountinfo` - log out again

3. Remove the workloads and persistent volumes

4. Shell in to the node again and list the mountpoints - see that "globalmounts" still remains even though the node should have been clean, and that the kubelet has begun logging errors - eg via `journalctl -u kubelet --output=short --since "5 minutes ago"`

Some additional things I've seen
* If I manually unmount one of the paired globalmounts, the the other gets unmounted by the driver without any problems
* I've only been able to reproduce this when two or more "unrelated" (that is, eg. not from the same Deployment) is scheduled to the sam node. If I only schedule a single pod I still get the double global-mounts but they are umounted without any problems.
* The fact that it is ever possible for globalmounts to get unmounted when they apparently come in pairs indicate that volumes normally gets unmounted before the kublet gets to do it's first `GetDeviceMountRefs` - in such as situation volumes would get unmounted just fine as long as the processes succeeds in the first try. But, if we ever reach the point where the kubelet does a GetDeviceMountRefs - we'll never get out.

**Environment**:
- CSI Driver version: `1.23.0`
- Kubernetes version: `v1.24.6` (I've tested with multiple patch versions of 1.23 as well)
- OS: `Ubuntu 18.04.6 LTS (Bionic Beaver)`
- Kernel: `Linux aks-agentpool-16100435-vmss000002 5.4.0-1094-azure #100~18.04.1-Ubuntu SMP Mon Oct 17 11:44:30 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux`

danquah:
**What type of PR is this?**

/kind documentation

Improves a comment for volumeHandle and adds the `nosharesock` mount option for all `PersistentVolume` examples.

This is done to better support the situation where the creator of a PersistentVolume wants to use the same Azure Files share multiple times.

Fixes #1137

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/deploy/example/nginx-pod-azurefile-inline-volume.yaml
+++ b/deploy/example/nginx-pod-azurefile-inline-volume.yaml
@@ -23,4 +23,4 @@ spec:
         volumeAttributes:
           shareName: EXISTING_SHARE_NAME  # required
           secretName: azure-secret  # required
-          mountOptions: "dir_mode=0777,file_mode=0777,cache=strict,actimeo=30"  # optional
+          mountOptions: "dir_mode=0777,file_mode=0777,cache=strict,actimeo=30,nosharesock"  # optional
--- a/deploy/example/pv-azurefile-csi.yaml
+++ b/deploy/example/pv-azurefile-csi.yaml
@@ -21,7 +21,7 @@ spec:
   csi:
     driver: file.csi.azure.com
     readOnly: false
-    # make sure this volumeid is unique in the cluster
+    # make sure volumeid is unique for every identical share in the cluster
     # `#` is not allowed in self defined volumeHandle
     volumeHandle: unique-volumeid
     volumeAttributes:
--- a/deploy/example/pv-azurefile-nfs.yaml
+++ b/deploy/example/pv-azurefile-nfs.yaml
@@ -12,7 +12,7 @@ spec:
   csi:
     driver: file.csi.azure.com
     readOnly: false
-    # make sure this volumeid is unique in the cluster
+    # make sure volumeid is unique for every identical share in the cluster
     # `#` is not allowed in self defined volumeHandle
     volumeHandle: unique-volumeid
     volumeAttributes:
--- a/docs/intree-static-example.md
+++ b/docs/intree-static-example.md
@@ -147,6 +147,7 @@ spec:
   - gid=1000
   - mfsymlinks
   - nobrl
+  - nosharesock
 ```
 
 To update your mount options, create a *azurefile-mount-options-pv.yaml* file with a *PersistentVolume*. For example:
@@ -172,6 +173,7 @@ spec:
   - gid=1000
   - mfsymlinks
   - nobrl
+  - nosharesock
 ```
 
 Create a *azurefile-mount-options-pvc.yaml* file with a *PersistentVolumeClaim* that uses the *PersistentVolume*. For example:


andyzhangx:
@danquah thanks for contribution, the commit message is invalid, pls run `git commit --amend` to rephrase the commit msg, and the run "git push ... --force" to overwrite original commit, thanks.

danquah:
Done.

Would it be an idea to add some more documentation on how to use volumeHandle? https://github.com/kubernetes-sigs/azurefile-csi-driver/blob/master/docs/driver-parameters.md already mentions the `volumeHandle` generated for dynamic volumes:

> - VolumeID(`volumeHandle`) is the identifier of the volume handled by the driver, format of VolumeID: 
> ```
> {resource-group-name}#{account-name}#{file-share-name}#{placeholder}#{uuid}#{secret-namespace}
> ```
> > `placeholder`, `uuid`, `secret-namespace` are optional

A new section could be added under [Static Provision](https://github.com/kubernetes-sigs/azurefile-csi-driver/blob/master/docs/driver-parameters.md#static-provisionbring-your-own-file-share) that has recommendations for how to pick a value for `volumeHandle`. As storage account names [are globally unique](https://learn.microsoft.com/en-us/azure/storage/common/storage-account-overview#storage-account-name) I think the following could be enough

> Specify a value for `volumeHandle` that is  unique for every identical share in the cluster. A recommended way to produce such as value is to combine the Azure globally unique name storage account name and the name of the share:
> ```
> {account-name}_{file-share-name}
> ```
> Example:
> ```
> supercorp2000account_department-123-share
> ```
>  > Be aware the character `#` cannot is reserved for internal use and cannot be used in a volumeHandle

andyzhangx (approved):
/lgtm
/retest

danquah:
is the test flaky or are any of the example files actually used in a test?

andyzhangx:
> Done.
> 
> Would it be an idea to add some more documentation on how to use volumeHandle? https://github.com/kubernetes-sigs/azurefile-csi-driver/blob/master/docs/driver-parameters.md already mentions the `volumeHandle` generated for dynamic volumes:
> 
> > * VolumeID(`volumeHandle`) is the identifier of the volume handled by the driver, format of VolumeID:
> > 
> > ```
> > {resource-group-name}#{account-name}#{file-share-name}#{placeholder}#{uuid}#{secret-namespace}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > > `placeholder`, `uuid`, `secret-namespace` are optional
> 
> A new section could be added under [Static Provision](https://github.com/kubernetes-sigs/azurefile-csi-driver/blob/master/docs/driver-parameters.md#static-provisionbring-your-own-file-share) that has recommendations for how to pick a value for `volumeHandle`. As storage account names [are globally unique](https://learn.microsoft.com/en-us/azure/storage/common/storage-account-overview#storage-account-name) I think the following could be enough
> 
> > Specify a value for `volumeHandle` that is  unique for every identical share in the cluster. A recommended way to produce such as value is to combine the Azure globally unique name storage account name and the name of the share:
> > ```
> > {account-name}_{file-share-name}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Example:
> > ```
> > supercorp2000account_department-123-share
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > > Be aware the character `#` cannot is reserved for internal use and cannot be used in a volumeHandle

@danquah I think that's reaonsable, just go ahead submitting a PR, thanks.

PR #1156, opened 2023-01-04
fix: account matching issue in account search

### LINKED ISSUE #1154: Same storage account for different requireInfraEncryption values
**What happened**:

Having two StorageClasses, each with different `requireInfraEncryption: "{true|false}"`, the driver creates Storage Account only for one of them (the one used the first) and leaves all PVs either encrypted or unencrypted at rest.

Two Storage Accounts, one for encrypted and second for unencrypted volumes. Or encryption scope used for the encrypted ones.

**How to reproduce it**:

1. Create a SC with `requireInfraEncryption: "false"` + a PVC:

```
kubectl apply -f - <<EOF
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: test
parameters:
  skuName: Standard_LRS
  requireInfraEncryption: "false"

---

kind: PersistentVolumeClaim
apiVersion: v1
metadata:
  name: myclaim
spec:
  storageClassName: test
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 3500Mi
EOF
```

2. Create a second SC with `requireInfraEncryption: "true"` and a PVC:
```
kubectl apply -f - <<EOF
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: test2
parameters:
  skuName: Standard_LRS
  requireInfraEncryption: "true"

---

kind: PersistentVolumeClaim
apiVersion: v1
metadata:
  name: myclaim2
spec:
  storageClassName: test2
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 3500Mi
EOF
```

3. Check Azure Storage Accounts, there is only one and it has "Infrastructure encryption: Disabled" in UI.

**Environment**:

- CSI Driver version: 1.21.0
- Kubernetes version (use `kubectl version`):  v1.25.2

andyzhangx:
**What type of PR is this?**
/kind bug

fix: account matching issue in account search

Fixes #1154

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
fix: account matching issue in account search
```


jsafrane:
Thanks for looking into this!

I think the vendor changes should include `AllowSharedKeyAccess` + `IsMultichannelEnabled` properties, even when this CSI driver won't use them. There may be other users of Azure SDK that could benefit from such a bugfix.

jsafrane:
/lgtm
Thanks a lot!

PR #1193, opened 2023-02-16
feat: Support workload identity

### LINKED ISSUE #1138: Add support for Workload Identity

 - A customer using ARO wants to spin up an OpenShift cluster with "az aro create" without needing additional input, i.e. without the need for an AD account or service principal credentials, and the identity used is never visible to the customer and cannot appear in the cluster.
 
 - As an administrator, I want to deploy OpenShift 4 and run Operators on Azure using access controls (IAM roles) with temporary, limited privilege credentials.

CSI Driver should support identity federation to enable workload identities, similar work is currently WIP in [cluster-api-provider-azure](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2814/files#diff-8f168651958e0a88d9402ac98e3b8764c9a4b54d29de8284def5e2c9070dd11bR280).

In order to use identity federation the driver has to be updated to use azidentity instead of autorest/adal.

Alternatively, if we can't migrate the driver to azidentity, OpenShift Operators might use OIDC sidecar as a short term solution: https://learn.microsoft.com/en-us/azure/aks/workload-identity-overview#how-to-migrate-to-workload-identity

**Additional context**

OpenShift Enhancement: https://github.com/openshift/enhancements/blob/40aac25619eb2e1bd2fb55a90bdcdddf7d7346e2/enhancements/cloud-integration/azure/azure-workload-identity.md

cvvz:
**What type of PR is this?**
/kind feature

Fixes #1138

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


cvvz:
workload identity failed to inject to csi driver when csi driver was deployed in `kube-system` namespace. @karataliu is confirming.

charts/latest/azurefile-csi-driver/templates/csi-azurefile-controller.yaml
```diff
@@ -193,6 +193,16 @@ spec:
...
+            - name: AZURE_CLIENT_ID
+              value: {{ .Values.workloadIdentity.clientID }}
+            - name: AZURE_TENANT_ID
+              value: {{ .Values.workloadIdentity.tenantID }}
+            - name: AZURE_FEDERATED_TOKEN_FILE
+              value: /var/run/secrets/tokens/azure-identity-token
+            - name: AZURE_AUTHORITY_HOST
+              value: https://login.microsoftonline.com/
+{{- end }}
             - name: AZURE_CREDENTIAL_FILE
```
primeroz, line 206: `else` AZURE_CREDENTIAL_FILE ?

charts/latest/azurefile-csi-driver/templates/csi-azurefile-controller.yaml
```diff
@@ -193,6 +193,16 @@ spec:
             timeoutSeconds: 10
             periodSeconds: 30
           env:
+{{- if and .Values.workloadIdentity.clientID .Values.workloadIdentity.tenantID}}
```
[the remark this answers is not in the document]
cvvz: We won't rely on Azure Workload Identity Webhook to inject the environment variables

charts/latest/azurefile-csi-driver/templates/csi-azurefile-controller.yaml
primeroz: :+1:

charts/latest/azurefile-csi-driver/templates/csi-azurefile-controller.yaml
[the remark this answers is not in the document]
primeroz: :+1:

charts/latest/azurefile-csi-driver/templates/csi-azurefile-controller.yaml
```diff
@@ -227,6 +242,16 @@ spec:
...
           resources: {{- toYaml .Values.controller.resources.azurefile | nindent 12 }}
       volumes:
+{{- if and .Values.workloadIdentity.clientID .Values.workloadIdentity.tenantID}}
+        - name: azure-identity-token
+          projected:
+            defaultMode: 420
+            sources:
+            - serviceAccountToken:
+                audience: api://AzureADTokenExchange
+                expirationSeconds: 3600
```
andyzhangx, line 252: need to check when token expire, how does this driver do refresh token? is it automatic?

charts/latest/azurefile-csi-driver/templates/csi-azurefile-controller.yaml
```diff
@@ -193,6 +193,16 @@ spec:
...
           env:
+{{- if and .Values.workloadIdentity.clientID .Values.workloadIdentity.tenantID}}
+            - name: AZURE_CLIENT_ID
+              value: {{ .Values.workloadIdentity.clientID }}
+            - name: AZURE_TENANT_ID
+              value: {{ .Values.workloadIdentity.tenantID }}
+            - name: AZURE_FEDERATED_TOKEN_FILE
+              value: /var/run/secrets/tokens/azure-identity-token
+            - name: AZURE_AUTHORITY_HOST
+              value: https://login.microsoftonline.com/
```
andyzhangx, line 204: this is the global setting, what if not set? and also need to support sovereign cloud.

andyzhangx:
cc @RomanBednar for this PR, let me know if you have any concern.

andyzhangx:
also we need to check how to enable workload identity on capz, there is a draft PR to support workload identity on CAPZ: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2814

charts/latest/azurefile-csi-driver/templates/csi-azurefile-controller.yaml
cvvz: According to the [official doc](https://learn.microsoft.com/en-us/azure/aks/workload-identity-overview#:~:text=It%27s%20an%20optional%20field%20that%20you%20configure%20to%20prevent%20downtime%0Acaused%20by%20errors%20during%20service%20account%20token%20refresh.%20Kubernetes%20service%20account%20token%20expiry%20isn%27t%20correlated%20with%20Azure%20AD%20tokens.%20Azure%20AD%20tokens%20expire%20in%2024%20hours%20after%20they%27re%20issued.), it will be automatically refreshed. Application Pod don't need to do anything.

andyzhangx: but we are using old adal lib, will that token be refreshed after 24 hrs?

cvvz: You are right, although the token was refreshed, the client won't refresh its Authorizer header and encountered authentication error, I will take a look at it.

cvvz: adal lib do have the mechanism that will refresh the token after it expired. However there is a bug that will trigger this error under certain circumstances. I've filed a [PR](https://github.com/Azure/go-autorest/pull/719) to fix it, hopefully it can be merged before end-of-life.

charts/latest/azurefile-csi-driver/templates/csi-azurefile-controller.yaml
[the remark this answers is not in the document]
abutcher: :+1: We won't rely on the webhook either because we [set the env vars on the azurefile deployment as read from a secret](https://github.com/openshift/azure-file-csi-driver-operator/blob/994c32c927a69df33f83aed440cc150e8d270341/assets/controller.yaml#L54-L64). I think this will align with our implementation since we can just extend this to set `AZURE_FEDERATED_TOKEN_FILE` & `AZURE_TENANT_ID` which will mean that the workload identity extension will be used instead of service principal.

RomanBednar:
> cc @RomanBednar for this PR, let me know if you have any concern.

Thank you @andyzhangx and @cvvz for the PR. The proposed changes look great for the purposes of our storage operator I believe. We don't use helm to deploy the driver but we can implement something similar.

RomanBednar:
@cvvz I gave it a bit more thought and believe it would be much cleaner for deployments and enabling/disabling workload identity authentication feature if the auth lib could also decide about the method used based on cloud config. Keeping the env variables for the sake of webhook is still ok - we could have both.

It makes sense for us in OpenShift enhancement but I'd like you to weight in on that. Here's the enhancement discussion: https://github.com/openshift/enhancements/pull/1301#discussion_r1135195873

cvvz:
@RomanBednar It seems good to me to support federated workload identity in cloud auth config, which is a CCO scope configuration. And we should still keep using the env variables and webhook in csi driver, since some users may want to configure workload identity separately.

cvvz:
Hi, @RomanBednar I just checked again, and I think we already support workload identity based on cloud config in this PR. 

You can set `aadClientId`, `tenantId`, `aadFederatedTokenFile`, and `useFederatedWorkloadIdentityExtension` in auth config file, which will eventually be parsed as [AzureAuthConfig](https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/pkg/provider/config/azure_auth.go#L50-L81), and don't set any env, so the configuration won't be overwrited.

RomanBednar:
> You can set `aadClientId`, `tenantId`, `aadFederatedTokenFile`, and `useFederatedWorkloadIdentityExtension` in auth config file

Ah I see, initially I misunderstood the config parsing, thanks for pointing that out! @cvvz

charts/latest/azurefile-csi-driver/values.yaml
```diff
@@ -253,3 +253,9 @@ windows:
               operator: NotIn
               values:
                 - virtual-kubelet
+
+workloadIdentity:
```
andyzhangx, line 257: pls also add parameter description into https://github.com/kubernetes-sigs/azurefile-csi-driver/tree/master/charts

andyzhangx (approved):
pls squash all commits, thanks.

PR #1202, opened 2023-03-09
[WIP]vendor: pick up a random matching account

andyzhangx:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/vendor/sigs.k8s.io/cloud-provider-azure/pkg/provider/azure_storageaccount.go
+++ b/vendor/sigs.k8s.io/cloud-provider-azure/pkg/provider/azure_storageaccount.go
@@ -19,6 +19,7 @@ package provider
 import (
 	"context"
 	"fmt"
+	"math/rand"
 	"strings"
 
 	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2022-07-01/network"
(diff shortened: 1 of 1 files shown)


vendor/sigs.k8s.io/cloud-provider-azure/pkg/provider/azure_storageaccount.go
```diff
@@ -212,7 +214,9 @@ func (az *Cloud) EnsureStorageAccount(ctx context.Context, accountOptions *Accou
 			}
 
 			if len(accounts) > 0 {
-				accountName = accounts[0].Name
+				// pick up a random one of matching accounts
+				rand.Seed(time.Now().UnixNano())
+				accountName = accounts[rand.Intn(len(accounts))].Name
```
jsafrane, line 219: I am not sure if you are going to actually propose a random account selection, I would definitely appreciate if the storage account was chosen deterministically - e.g. the oldest one / the newest one / the first in alphabet... Randomness will bite us in the long run.

andyzhangx: hi @jsafrane it's probably yes since there are users hit egress limit for one storage account, this PR is just for POC. We will probably create a new parameter for this new algorithm, by default, it would still pick up the first matching account. And if users could change the account selection behavior if they want random.

PR #1363, opened 2023-08-14
feat(helm) Allow setting an alternate name of the configmap for the azure credential path

mtougeron:
**What type of PR is this?**

/kind feature

This allows setting an name for the azure credential configmap. This allows this chart and the `azuredisk-csi-driver` chart to not conflict with each other over the name of the configmap (associated PR https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/1935). This is important for a GitOps workflow where we don't want an order of operations dependency between these two charts.

We install these charts via a wrapper chart that installs the additional configmap manifest with the appropriate config information. However, both charts cannot contain that extra manifest or Argo CD would create warnings that more than 1 chart owns the configmap. This means that the azuredisk and azurefile wrapper charts would need to install the configmap with different names.

Otherwise, if only one of the two wrapper charts installed the additional manifest, there would be an order of operations dependency on the two charts. This creates additional complexity that Argo CD does not have a mechanism to handle.

Fixes # n/a

**Requirements**:
- [x] includes documentation
- [n/a] adds unit tests

**Release note**:
```
Allow setting an alternate name of the configmap for the azure credential path in the helm chart
```


adriananeci (commented):
/lgtm

mtougeron:
flake on the test `net/http: TLS handshake timeout`

/test pull-azurefile-csi-driver-e2e-migration-windows

mtougeron:
I'm not seeing anything related to my PR in these errors. The manifest is rendering properly as expected. Going to retest.

/retest-required

andyzhangx (approved):
/lgtm

PR #1634, opened 2023-12-21
Refactor: remove grpc wrapper

MartinForReal:
**What type of PR is this?**
/kind cleanup

Refactor: remove grpc wrapper

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [x] adds unit tests

**Release note**:
```
none
```


pkg/azurefile/azurefile.go
```diff
@@ -407,10 +410,29 @@ func (d *Driver) Run(endpoint, kubeconfig string, testBool bool) {
 	}
 	d.AddNodeServiceCapabilities(nodeCap)
 
-	s := csicommon.NewNonBlockingGRPCServer()
-	// Driver d act as IdentityServer, ControllerServer and NodeServer
-	s.Start(endpoint, d, d, d, testBool)
-	s.Wait()
+	//setup grpc server
+	opts := []grpc.ServerOption{
```
andyzhangx, line 414: what's the intention for this?
why not wrap `NewNonBlockingGRPCServer` to make logic more clear?

MartinForReal: the wrapper is not shared among csi projects. 
It is only referred by main once. 
The wrapper doesn't introduce any extra function which grpc server provides.

MartinForReal: Also, we don't need to add any code for test purposes. the metrics related code can be put to one place. 
and the line of code decreased. 
I think this will make the code clearer.

pkg/azurefile/azurefile.go
```diff
@@ -407,10 +410,29 @@ func (d *Driver) Run(endpoint, kubeconfig string, testBool bool) {
...
+	listener, err := csicommon.ListenEndpoint(endpoint)
+	if err != nil {
+		klog.Fatalf("failed to listen endpoint: %v", err)
+	}
+	go func() {
+		<-ctx.Done()
+		d.server.GracefulStop()
+	}()
+	if err = d.server.Serve(listener); errors.Is(err, grpc.ErrServerStopped) {
+		klog.Infof("gRPC server stopped serving")
```
andyzhangx, line 432: why not return error here? for ut pass?

MartinForReal: If the context expired/closed, the grpc server will return ErrServerStopped, I don't think it is an error. Because it is an expected behavior

andyzhangx: ok, can you add a comment for this? and pls check whether it's possible to wrap logic from L413-L434 (similar to existing code), that would make logic more clear and make it possible to write ut for such code change.

andyzhangx: wait, this `Run` function should never stop,  and also don't need to return error since this func never stops

MartinForReal: yes. it will block until the server stops.

MartinForReal: I believe this is covered in pkg/azurefile/azurefile_test.go https://github.com/kubernetes-sigs/azurefile-csi-driver/blob/bd8e20bf5a8721088ed8fce0fceb17ab5d243617/pkg/azurefile/azurefile_test.go#L1019-L1106

andyzhangx (approved):
/lgtm

PR #1729, opened 2024-02-28
fix: ensure azure cloud config could be loaded from secret

jlpedrosa:
**What type of PR is this?**

/kind bug

Enables azure-csi to be run with the option to provide a cloud-config in a secret, which is required via the current helm chart for non AKS clusters.

Current implementation is shadowing the variables, so even on successful loading the config from the secret, the configuration is lost (nil) and ignored by the rest of the code.

**Requirements**:
- [X] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/azurefile/azure.go
+++ b/pkg/azurefile/azure.go
@@ -75,7 +75,7 @@ func getCloudProvider(ctx context.Context, kubeconfig, nodeID, secretName, secre
 
 	if kubeClient != nil {
 		klog.V(2).Infof("reading cloud config from secret %s/%s", secretNamespace, secretName)
-		config, err := configloader.Load[azure.Config](ctx, &configloader.K8sSecretLoaderConfig{
+		config, err = configloader.Load[azure.Config](ctx, &configloader.K8sSecretLoaderConfig{
 			K8sSecretConfig: configloader.K8sSecretConfig{
 				SecretName:      secretName,
 				SecretNamespace: secretNamespace,


jlpedrosa:
@andyzhangx: I see the tests are timing out, I do believe it's not related to this particular change. Should I use `retest` command until it works? I am assuming I'm allowed to do that once you indicated ok to thest?

andyzhangx:
> @andyzhangx: I see the tests are timing out, I do believe it's not related to this particular change. Should I use `retest` command until it works? I am assuming I'm allowed to do that once you indicated ok to thest?

@jlpedrosa just ignore this test since there is same failure in another PR, sth. wrong with the test suite to create a new vm.

andyzhangx (approved):
/lgtm

pkg/azurefile/azure.go
```diff
@@ -75,7 +75,7 @@ func getCloudProvider(ctx context.Context, kubeconfig, nodeID, secretName, secre
 
 	if kubeClient != nil {
 		klog.V(2).Infof("reading cloud config from secret %s/%s", secretNamespace, secretName)
-		config, err := configloader.Load[azure.Config](ctx, &configloader.K8sSecretLoaderConfig{
+		config, err = configloader.Load[azure.Config](ctx, &configloader.K8sSecretLoaderConfig{
```
andyzhangx, line 78: do you know whether we have any golint check or sth. else to avoid such issue? @MartinForReal @cvvz

PR #2048, opened 2024-08-07
chore: fix missing securityContext on controller pods

xrstf:
**What type of PR is this?**
/kind cleanup

In #1210 security contexts were added, but none was added to the snapshot controller and the CSI controller. Something similar happened in the azuredisk CSI repo, with the same oversight. In the azuredisk repo, the oversight was fixed in https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/1805 and this PR is the equivalent for the azurefile CSI.

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
Add securityContext to driver controller and snapshot controller Pods.
```


xrstf:
Did I accidentally actually break something?

andyzhangx:
> Did I accidentally actually break something?

@xrstf No, we are in the middle of migrating all GCP based tests to AWS, that would require efforts.

PR #2310, opened 2024-12-29
doc: Update examples for mount permissions setting

m-anyac:
**What type of PR is this?**
/kind documentation

Add comment for mount permission 
Fix installation guide

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
chore: add comment for mount permission and fix installation guide
```

### DIFF
--- a/deploy/example/pv-azurefile-csi.yaml
+++ b/deploy/example/pv-azurefile-csi.yaml
@@ -13,7 +13,7 @@ spec:
   persistentVolumeReclaimPolicy: Retain  # if set as "Delete" file share would be removed in pvc deletion
   storageClassName: azurefile-csi
   mountOptions:
-    - dir_mode=0777
+    - dir_mode=0777 # modify this permission if you want to enhance the security
     - file_mode=0777
     - uid=0
     - gid=0
--- a/deploy/example/pv-azurefile-csi.yaml
+++ b/deploy/example/pv-azurefile-csi.yaml
@@ -13,7 +13,7 @@ spec:
   persistentVolumeReclaimPolicy: Retain  # if set as "Delete" file share would be removed in pvc deletion
   storageClassName: azurefile-csi
   mountOptions:
-    - dir_mode=0777 # modify this permission if you want to enhance the security
+    - dir_mode=0777  # modify this permission if you want to enhance the security
     - file_mode=0777
     - uid=0
     - gid=0
--- a/deploy/example/storageclass-azurefile-csi.yaml
+++ b/deploy/example/storageclass-azurefile-csi.yaml
@@ -10,7 +10,7 @@ parameters:
 reclaimPolicy: Delete
 volumeBindingMode: Immediate
 mountOptions:
-  - dir_mode=0777 # modify this permission if you want to enhance the security
+  - dir_mode=0777  # modify this permission if you want to enhance the security
   - file_mode=0777
   - mfsymlinks
   - cache=strict  # https://linux.die.net/man/8/mount.cifs
--- a/deploy/example/storageclass-azurefile-csi.yaml
+++ b/deploy/example/storageclass-azurefile-csi.yaml
@@ -10,7 +10,7 @@ parameters:
 reclaimPolicy: Delete
 volumeBindingMode: Immediate
 mountOptions:
-  - dir_mode=0777
+  - dir_mode=0777 # modify this permission if you want to enhance the security
   - file_mode=0777
   - mfsymlinks
   - cache=strict  # https://linux.die.net/man/8/mount.cifs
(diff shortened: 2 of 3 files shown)


pkg/azurefile/azurefile_options.go
```diff
@@ -70,7 +70,7 @@ func (o *DriverOptions) AddFlags() *flag.FlagSet {
 	fs.BoolVar(&o.EnableGetVolumeStats, "enable-get-volume-stats", true, "allow GET_VOLUME_STATS on agent node")
 	fs.BoolVar(&o.EnableKataCCMount, "enable-kata-cc-mount", false, "enable Kata Confidential Containers mount")
 	fs.BoolVar(&o.AppendMountErrorHelpLink, "append-mount-error-help-link", true, "Whether to include a link for help with mount errors when a mount error occurs.")
-	fs.Uint64Var(&o.MountPermissions, "mount-permissions", 0777, "mounted folder permissions")
+	fs.Uint64Var(&o.MountPermissions, "mount-permissions", 0775, "mounted folder permissions")
```
jsafrane, line 73: I think this is a breaking change. If I have an app that counts with 777 as the default and I update to this code, my application breaks. This needs to be announced properly, at least with a "breaking change" warning in the release notes.

I would personally also bump the major version, but I am not sure what are rules in this repo.

m-anyac: Hey Jan, thanks for pointing this out. It works on my end with 777. Could you provide more details about the crash, such as any error messages or related information?

nearora-msft: @edreed This could be breaking change as we are making the permissions more restrictive. Is it a hard requirement to move to 0775 as default? 
Should we just update the storageclass and other examples, or should we make this code change and mention in the release nodes so that the users are aware?

m-anyac: @jsafrane this will revert to 777 thanks

m-anyac:
review-requested:andyzhangx

m-anyac:
pr review-requested:andyzhangx

m-anyac:
Hi Andy @andyzhangx, could I get review from you? 

I think I have no permission to add a reviewer to this PR.
Screenshot 2025-01-06 185128

andyzhangx:
@m-anyac 
if you want to make more secure, then why not use `0755` ? and I don't think you need to change all examples, you only need to add a comment after 0777, e.g. `0777 # modify this permission if you want to enhance the security`

nearora-msft:
> @m-anyac if you want to make more secure, then why not use `0755` ? and I don't think you need to change all examples, you only need to add a comment after 0777, e.g. `0777 # modify this permission if you want to enhance the security`

@andyzhangx Isn't a customer is more likely to use the example directly, instead of modifying the example based on comments? If we want this setting to be default in the future, it makes sense to modify the examples first?

andyzhangx:
> > @m-anyac if you want to make more secure, then why not use `0755` ? and I don't think you need to change all examples, you only need to add a comment after 0777, e.g. `0777 # modify this permission if you want to enhance the security`
> 
> @andyzhangx Isn't a customer is more likely to use the example directly, instead of modifying the example based on comments? If we want this setting to be default in the future, it makes sense to modify the examples first?

@nearora-msft different customers want different settings, in the before, we have set as 0755 by default in the code, and then challenged by quite a lot customers since default setting won't work for them, and finally we reverted to 0777. I think we should set 0777 for general usage, and if specific customers wants to be more secure, then they could use 0775 or 0755, there is no setting applied for all customers, that's not possible.

andyzhangx (approved):
/lgtm

deploy/example/pv-azurefile-csi.yaml
```diff
@@ -13,7 +13,7 @@ spec:
   persistentVolumeReclaimPolicy: Retain  # if set as "Delete" file share would be removed in pvc deletion
   storageClassName: azurefile-csi
   mountOptions:
-    - dir_mode=0777
+    - dir_mode=0777 # modify this permission if you want to enhance the security
```
andyzhangx, line 16:
```
deploy/example/pv-azurefile-csi.yaml:16:21: [warning] too few spaces before comment (comments)
deploy/example/storageclass-azurefile-csi.yaml:13:19: [warning] too few spaces before comment (comments)
```

pls insert another space before `#`

m-anyac: thanks for helping

andyzhangx:
Ubuntu apt update is down, so this failure is expected now:
```
7.898 Processing triggers for libc-bin (2.36-9+deb12u8) ...
7.913 Processing triggers for ca-certificates (20230311) ...
7.919 Updating certificates in /etc/ssl/certs...
8.736 0 added, 0 removed; done.
8.736 Running hooks in /etc/ca-certificates/update.d...
8.738 done.
9.091 
9.091 gzip: stdin: not in gzip format
9.092 tar: Child returned status 1
9.092 tar: Error is not recoverable: exiting now
```

PR #2314, opened 2025-01-03
chore: set GoMaxProcs as 2

andyzhangx:
**What type of PR is this?**
/kind feature

chore: set GoMaxProcs as 2

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/azurefile/azurefile_options.go
+++ b/pkg/azurefile/azurefile_options.go
@@ -48,6 +48,7 @@ type DriverOptions struct {
 	PrintVolumeStatsCallLogs               bool
 	SasTokenExpirationMinutes              int
 	WaitForAzCopyTimeoutMinutes            int
+	GoMaxProcs                             int
 	KubeConfig                             string
 	Endpoint                               string
 }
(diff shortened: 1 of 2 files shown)


pkg/azurefile/azurefile_options.go
```diff
@@ -87,6 +88,7 @@ func (o *DriverOptions) AddFlags() *flag.FlagSet {
 	fs.IntVar(&o.WaitForAzCopyTimeoutMinutes, "wait-for-azcopy-timeout-minutes", 19, "timeout in minutes for waiting for azcopy to finish")
 	fs.StringVar(&o.KubeConfig, "kubeconfig", "", "Absolute path to the kubeconfig file. Required only when running out of cluster.")
 	fs.StringVar(&o.Endpoint, "endpoint", "unix://tmp/csi.sock", "CSI endpoint")
+	fs.IntVar(&o.GoMaxProcs, "max-procs", 2, "maximum number of CPUs that can be executing simultaneously in golang runtime")
```
MartinForReal, line 91: https://github.com/uber-go/automaxprocs/blob/master/maxprocs/maxprocs.go the process will be throttled. I think it is better to round the quota based on the config?

MartinForReal: but maxprocs may have compatibility issue when it is used with cgroups

andyzhangx: it's using 2 CPU at most, see detailed explanation here: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/968, google driver has set as 1

Issue #2423, opened 2025-03-10
Resizing related tests broken with latest version


gnufied:
I am as well debugging, but it looks like volume resizing tests are broken with latest version (at least with Openshift), when using driver with NFS volumes:

```
833f997adf6a/resourceGroups/ci-op-1qvf1w3y-64bff-h557n-rg/providers/Microsoft.Storage/storageAccounts/f54f2cae4810d480e8921c6/fileServices/default/shares/pvcn-4880d75c-fa96-4745-9130-41bd8729e6a6
  --------------------------------------------------------------------------------
  RESPONSE 400: 400 Bad Request
  ERROR CODE: InvalidHeaderValue
  --------------------------------------------------------------------------------
  {
    "error": {
      "code": "InvalidHeaderValue",
      "message": "The value for one of the HTTP headers is not in the correct format.\nRequestId:ed74c408-901a-009c-5fe0-8fba34000000\nTime:2025-03-08T04:11:25.3570572Z"
    }
  }
  --------------------------------------------------------------------------------

  I0308 04:21:30.300783 1892 dump.go:53] At 2025-03-08 04:11:25 +0000 UTC - event for file.csi.azure.comnrr48: {external-resizer file.csi.azure.com } VolumeResizeFailed: resize volume "pvc-4880d75c-fa96-4745-9130-41bd8729e6a6" by resizer "file.csi.azure.com" failed: rpc error: code = Internal desc = expand volume error: PATCH https://management.azure.com/subscriptions/d38f1e38-4bed-438e-b227-833f997adf6a/resourceGroups/ci-op-1qvf1w3y-64bff-h557n-rg/providers/Microsoft.Storage/storageAccounts/f54f2cae4810d480e8921c6/fileServices/default/shares/pvcn-4880d75c-fa96-4745-9130-41bd8729e6a6
  --------------------------------------------------------------------------------
  RESPONSE 400: 400 Bad Request
  ERROR CODE: InvalidHeaderValue
  --------------------------------------------------------------------------------
  {
    "error": {
      "code": "InvalidHeaderValue",
      "message": "The value for one of the HTTP headers is not in the correct format.\nRequestId:ed74c430-901a-009c-02e0-8fba34000000\nTime:2025-03-08T04:11:25.6306763Z"
    }
}
```

[...]

PR #2431, opened 2025-03-17
Fix volume expansion for larger sizes

gnufied:
Fixes https://github.com/kubernetes-sigs/azurefile-csi-driver/issues/2423

### DIFF
--- a/pkg/azurefile/azurefile.go
+++ b/pkg/azurefile/azurefile.go
@@ -1032,6 +1032,11 @@ func (d *Driver) ResizeFileShare(ctx context.Context, subsID, resourceGroup, acc
 			if err != nil {
 				return true, err
 			}
+			if *fileShare.FileShareProperties.ShareQuota >= int32(sizeGiB) {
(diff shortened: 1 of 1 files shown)


pkg/azurefile/azurefile.go
```diff
@@ -1032,12 +1033,14 @@ func (d *Driver) ResizeFileShare(ctx context.Context, subsID, resourceGroup, acc
 			if err != nil {
 				return true, err
 			}
+			if ptr.Deref(fileShare.FileShareProperties.ShareQuota, 0) >= int32(sizeGiB) {
+				klog.Warningf("file share size(%dGi) is already greater or equal than requested size(%dGi), accountName: %s, shareName: %s",
```
gnufied, line 1037: so, if `fileShare.FileShareProperties.ShareQuota` was nil, then IMO the logging statement below should have also guarded against printing nil pointer. But that is perhaps a minor nit. :/

andyzhangx: https://github.com/kubernetes-sigs/azurefile-csi-driver/pull/2433 would fix the issue, thx

PR #2600, opened 2025-06-05
fix: only use the permission bits for chmod

### LINKED ISSUE #2606: Incorrect comparison of volumeattributes.mountPermissions mode bits with the targetPath permission bits
**What happened**:

The issue is that when the nfs fileshare is supposed to be mounted to another node, we reapply the gid settings in the SetVolumeOnwership, but we shouldn't do that because the gid settings were already applied the first time. But unfortunately, SetVolumeOnwership runs each time the fileshare needs to be mounted on a different node.

This happens when the chmodIfPermissionMismatch  ends up resetting the gid bit. Related issue: #2607.  And this happens when the user has volumeAttributes.mountPermissions set apart from the fsgroup settings.

image

When mountPermissions are set, we check if the mount permissions on the current nfs target path are equivalent to the permissions specified in the volumeAttributes.mountPermissions. 

Let's consider the case where the user's mountPermissions are set to "02775". As per the current code we are comparing info.Mode() & os.ModePerm with mode, which means we end up comparing "775" with "02775", which leads to a mismatch. This happens because we are comparing the permission bits of the nfs target path with all the mode bits set of the mountPermissions specified by the user.

Compare the mode bits from the target path and mountPermissions correctly

**How to reproduce it**:

**Environment**:

- CSI Driver version:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

nearora-msft:
**What type of PR is this?**
/kind bug

If the volumeattributes.mountpermissions are set for an nfs volume, we check if the permissions on the mount targetPath are the same as the expected mountPermissions. When this check is made currently, the targetPath mode bits are masked to use only the permission bits.
We should do the same with the mountPermissions bits to ensure that there isn't an unnecessary mismatch.
After the comparison, if the mount permissions are different, this PR ensures that only mount permissions change with the chmod operation and the rest of the bits like gid remain as is. 

Fixes #2606

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/azurefile/utils.go
+++ b/pkg/azurefile/utils.go
@@ -244,9 +244,11 @@ func chmodIfPermissionMismatch(targetPath string, mode os.FileMode) error {
 		return err
 	}
 	perm := info.Mode() & os.ModePerm
-	if perm != mode {
-		klog.V(2).Infof("chmod targetPath(%s, mode:0%o) with permissions(0%o)", targetPath, info.Mode(), mode)
-		if err := os.Chmod(targetPath, mode); err != nil {
+	expectedPerms := mode & os.ModePerm
+	if perm != expectedPerms {
+		klog.V(2).Infof("chmod targetPath(%s, mode:0%o) with permissions(0%o)", targetPath, info.Mode(), expectedPerms)
+		// only change the permission mode bits, keep the other bits as is
+		if err := os.Chmod(targetPath, (info.Mode()&^os.ModePerm)|os.FileMode(expectedPerms)); err != nil {
 			return err
 		}
 	} else {
--- a/pkg/azurefile/utils_test.go
+++ b/pkg/azurefile/utils_test.go
@@ -440,11 +440,23 @@ func TestChmodIfPermissionMismatch(t *testing.T) {
 	_ = makeDir(permissionMismatchPath, 0721)
 	defer os.RemoveAll(permissionMismatchPath)
 
+	permissionMatchGidMismatchPath, _ := getWorkDirPath("permissionMatchGidMismatchPath")
+	_ = makeDir(permissionMatchGidMismatchPath, 0755)
+	_ = os.Chmod(permissionMatchGidMismatchPath, 0755|os.ModeSetgid) // Setgid bit is set
+	defer os.RemoveAll(permissionMatchGidMismatchPath)
+
+	permissionMismatchGidMismatch, _ := getWorkDirPath("permissionMismatchGidMismatch")
+	_ = makeDir(permissionMismatchGidMismatch, 0721)
+	_ = os.Chmod(permissionMismatchGidMismatch, 0721|os.ModeSetgid) // Setgid bit is set
+	defer os.RemoveAll(permissionMismatchGidMismatch)
+
 	tests := []struct {
-		desc          string
-		path          string
-		mode          os.FileMode
-		expectedError error
+		desc           string
+		path           string
+		mode           os.FileMode
+		expectedPerms  os.FileMode
+		expectedGidBit bool
+		expectedError  error
 	}{
 		{
 			desc:          "Invalid path",
@@ -453,16 +465,44 @@ func TestChmodIfPermissionMismatch(t *testing.T) {
 			expectedError: fmt.Errorf("CreateFile invalid-path: The system cannot find the file specified"),
 		},
 		{
-			desc:          "permission matching path",
-			path:          permissionMatchingPath,
-			mode:          0755,
-			expectedError: nil,
+			desc:           "permission matching path",
+			path:           permissionMatchingPath,
+			mode:           0755,
+			expectedPerms:  0755,
+			expectedGidBit: false,
+			expectedError:  nil,
 		},
 		{
-			desc:          "permission mismatch path",
-			path:          permissionMismatchPath,
-			mode:          0755,
-			expectedError: nil,
+			desc:           "permission mismatch path",
+			path:           permissionMismatchPath,
+			mode:           0755,
+			expectedPerms:  0755,
+			expectedGidBit: false,
+			expectedError:  nil,
+		},
+		{
+			desc:           "only match the permission mode bits",
+			path:           permissionMatchGidMismatchPath,
+			mode:           0755,
+			expectedPerms:  0755,
+			expectedGidBit: true,
+			expectedError:  nil,
+		},
+		{
+			desc:           "only change the permission mode bits when gid is set",
+			path:           permissionMismatchGidMismatch,
+			mode:           0755,
+			expectedPerms:  0755,
+			expectedGidBit: true,
+			expectedError:  nil,
+		},
+		{
+			desc:           "only change the permission mode bits when gid is not set but mode bits have gid set",
+			path:           permissionMismatchPath,
+			mode:           02755,
+			expectedPerms:  0755,
+			expectedGidBit: false,
+			expectedError:  nil,
 		},
 	}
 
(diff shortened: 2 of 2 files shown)


pkg/azurefile/utils.go
```diff
@@ -244,9 +244,10 @@ func chmodIfPermissionMismatch(targetPath string, mode os.FileMode) error {
 		return err
 	}
 	perm := info.Mode() & os.ModePerm
-	if perm != mode {
-		klog.V(2).Infof("chmod targetPath(%s, mode:0%o) with permissions(0%o)", targetPath, info.Mode(), mode)
-		if err := os.Chmod(targetPath, mode); err != nil {
+	expectedPerms := mode & os.ModePerm
+	if perm != expectedPerms {
```
andyzhangx, line 248: the problem is not here, it's calling `SetVolumeOwnership`, and could take lots of time when there are large amount of small files: https://github.com/kubernetes-sigs/azurefile-csi-driver/blob/f368f9fdaf615c8ffc42156a727946b3f931fd5b/pkg/azurefile/nodeserver.go#L518-L524

nearora-msft: I understand that.
The issue is that when the nfs fileshare is supposed to be mounted to another node, we reapply the gid settings in the SetVolumeOnwership, but we shouldn't do that because the gid settings were already applied the first time. But unfortunately, SetVolumeOnwership runs each time the fileshare needs to be mounted on a different node.

This happens because the chmod that we do before in the code where I've made the changes, ends up resetting the gid bit. And this happens when the user has volumeAttributes.mountPermissions set apart from the fsgroup settings.

image

When mountPermissions are set, we check if the mount permissions on the current nfs target path are equivalent to the permissions specified in the volumeAttributes.mountPermissions. 

Let's consider the case of the ICM where the user's mountPermissions are set to "02775". As perr the current code we are comparing info.Mode() & os.ModePerm with mode, which means we end up comparing "775" with "02775", which leads to a mismatch.

nearora-msft: When this mismatch happens, we end up running os.chmod(targetPath, "02775"), which ends up resetting the gid bit of the targetPath. When chmod is done on the nfs target path with these permissions, the mode bits change from “20020000775” to “20000000775”

As a result of which when we check if we should skip the SetVolumeOwnership [here](https://github.com/kubernetes-sigs/azurefile-csi-driver/blob/9ff6b5b6047352a1a8411bbccdf42324d736fa86/vendor/k8s.io/kubernetes/pkg/volume/volume_linux.go#L53-L54), we find the gid bit is not set, so we should change the ownership recursively.

This could've been avoided if had compared the permission bits correctly and hand't run the chmod in the first place.

andyzhangx (approved):
/lgtm
/retest

andyzhangx:
@nearora-msft thanks, this looks reasonable, could you make similar change in blob csi driver for nfs volume mount? thx

andyzhangx (commented):
/hold
can you also add a ut in for `chmodIfPermissionMismatch` func?

nearora-msft:
Thanks for the review @andyzhangx , I would need to make a few more changes to this PR to accommodate how go handles chmod. But before that I have a few questions:
1. What is the purpose of having volumeAtributes.mountPermissions? 
2. If the gid bit is set in volumeAtributes.mountPermissions, example if volumeAtributes.mountPermissions=="02775", should we apply the gid bit in chmodIfPermissionmimatch, or should we wait for SetVolumeOwnership to do that?

andyzhangx:
> 2\. 02775



> Thanks for the review @andyzhangx , I would need to make a few more changes to this PR to accommodate how go handles chmod. But before that I have a few questions:
> 
> 1. What is the purpose of having volumeAtributes.mountPermissions?

`volumeAtributes.mountPermissions` is for setting the permission of root folder of nfs volume

> 2. If the gid bit is set in volumeAtributes.mountPermissions, example if volumeAtributes.mountPermissions=="02775", should we apply the gid bit in chmodIfPermissionmimatch, or should we wait for SetVolumeOwnership to do that?

we don't expect that gid bit is set in `volumeAtributes.mountPermissions`(it's `0777` by default), we only use SetVolumeOwnership to do that

nearora-msft:
> > 2. 02775
> 
> > Thanks for the review @andyzhangx , I would need to make a few more changes to this PR to accommodate how go handles chmod. But before that I have a few questions:
> > 
> > 1. What is the purpose of having volumeAtributes.mountPermissions?
> 
> `volumeAtributes.mountPermissions` is for setting the permission of root folder of nfs volume
> 
> > 2. If the gid bit is set in volumeAtributes.mountPermissions, example if volumeAtributes.mountPermissions=="02775", should we apply the gid bit in chmodIfPermissionmimatch, or should we wait for SetVolumeOwnership to do that?
> 
> we don't expect that gid bit is set in `volumeAtributes.mountPermissions`(it's `0777` by default), we only use SetVolumeOwnership to do that

Thanks @andyzhangx , I've updated the PR accordingly.

I'm thinking we should also prevent the user from setting gid bit in the mountPermissions, since it will never be picked up? 
We are checking the gid bit from the fsgroup settings in the security context, so it is a bit misleading for them if they set it in volumeAtributes.mountPermissions.

nearora-msft:
> If the volumeattributes.mountpermissions are set for an nfs volume, we check if the permissions on the mount targetPath are the same as the expected mountPermissions. When this check is made currently, the targetPath mode bits are masked to use only the permission bits.
> We should do the same with the mountPermissions bits to ensure that there isn't an unnecessary mismatch.



> @nearora-msft thanks, this looks reasonable, could you make similar change in blob csi driver for nfs volume mount? thx



> @nearora-msft thanks, this looks reasonable, could you make similar change in blob csi driver for nfs volume mount? thx

https://github.com/kubernetes-sigs/blob-csi-driver/pull/2041/files

andyzhangx:
@nearora-msft can you fix the windows ut failure? you could just skip running this test on Windows node since nfs volume is not supported on windows.
```
=== RUN   TestChmodIfPermissionMismatch
    utils_test.go:520: test[permission matching path]: unexpected perms: -rwxrwxrwx, expected perms: -rwxr-xr-x, 
    utils_test.go:520: test[permission mismatch path]: unexpected perms: -rwxrwxrwx, expected perms: -rwxr-xr-x, 
    utils_test.go:520: test[only match the permission mode bits]: unexpected perms: -rwxrwxrwx, expected perms: -rwxr-xr-x, 
    utils_test.go:524: test[only match the permission mode bits]: unexpected gid bit: false, expected gid bit: true
    utils_test.go:520: test[only change the permission mode bits when gid is set]: unexpected perms: -rwxrwxrwx, expected perms: -rwxr-xr-x, 
    utils_test.go:524: test[only change the permission mode bits when gid is set]: unexpected gid bit: false, expected gid bit: true
    utils_test.go:520: test[only change the permission mode bits when gid is not set but mode bits have gid set]: unexpected perms: -rwxrwxrwx, expected perms: -rwxr-xr-x, 
--- FAIL: TestChmodIfPermissionMismatch (0.00s)
```

nearora-msft:
> @nearora-msft can you fix the windows ut failure? you could just skip running this test on Windows node since nfs volume is not supported on windows.
> 
> ```
> === RUN   TestChmodIfPermissionMismatch
>     utils_test.go:520: test[permission matching path]: unexpected perms: -rwxrwxrwx, expected perms: -rwxr-xr-x, 
>     utils_test.go:520: test[permission mismatch path]: unexpected perms: -rwxrwxrwx, expected perms: -rwxr-xr-x, 
>     utils_test.go:520: test[only match the permission mode bits]: unexpected perms: -rwxrwxrwx, expected perms: -rwxr-xr-x, 
>     utils_test.go:524: test[only match the permission mode bits]: unexpected gid bit: false, expected gid bit: true
>     utils_test.go:520: test[only change the permission mode bits when gid is set]: unexpected perms: -rwxrwxrwx, expected perms: -rwxr-xr-x, 
>     utils_test.go:524: test[only change the permission mode bits when gid is set]: unexpected gid bit: false, expected gid bit: true
>     utils_test.go:520: test[only change the permission mode bits when gid is not set but mode bits have gid set]: unexpected perms: -rwxrwxrwx, expected perms: -rwxr-xr-x, 
> --- FAIL: TestChmodIfPermissionMismatch (0.00s)
> ```

Done

andyzhangx (approved):
/lgtm
/hold cancel

PR #2652, opened 2025-07-05
[WIP] feat: support SMB mount with managed identity

andyzhangx:
**What type of PR is this?**
/kind feature

feat: support SMB mount with managed identity

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
feat: support SMB mount with managed identity
```

### DIFF
--- a/pkg/azurefile/nodeserver.go
+++ b/pkg/azurefile/nodeserver.go
@@ -263,8 +263,7 @@ func (d *Driver) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRe
 	// don't respect fsType from req.GetVolumeCapability().GetMount().GetFsType()
 	// since it's ext4 by default on Linux
 	var fsType, server, protocol, ephemeralVolMountOptions, storageEndpointSuffix, folderName string
-	var ephemeralVol bool
-	var encryptInTransit bool
+	var ephemeralVol, encryptInTransit, mountWithManagedIdentity bool
 	fileShareNameReplaceMap := map[string]string{}
 
 	mountPermissions := d.mountPermissions
@@ -298,7 +297,6 @@ func (d *Driver) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRe
 			fileShareNameReplaceMap[pvNameMetadata] = v
 		case mountPermissionsField:
 			if v != "" {
-				var err error
 				var perm uint64
 				if perm, err = strconv.ParseUint(v, 8, 32); err != nil {
 					return nil, status.Errorf(codes.InvalidArgument, "invalid mountPermissions %s", v)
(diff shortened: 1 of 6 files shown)


charts/latest/azurefile-csi-driver/templates/csi-azurefile-node.yaml
```diff
@@ -96,11 +96,18 @@ spec:
...
+              value: "{{ .Values.node.enableManagedIdentityAuth }}"
           volumeMounts:
             - name: host-usr
               mountPath: /host/usr
             - name: host-etc
               mountPath: /host/etc
+            {{- if .Values.node.enableManagedIdentityAuth }}
+            - name: mountpoint-dir
+              mountPath: /var/lib/kubelet/
+              mountPropagation: Bidirectional
```
mittachaitu, line 109: Do we need `mountPropagation` as Bidirectional? Can we update it as `hostToContainer`?

pkg/azurefile/nodeserver.go
```diff
@@ -401,18 +405,29 @@ func (d *Driver) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRe
...
-		}
-		if runtime.GOOS == "windows" {
-			mountOptions = []string{fmt.Sprintf("AZURE\\%s", accountName)}
-			sensitiveMountOptions = []string{accountKey}
+		if mountWithManagedIdentity && runtime.GOOS != "windows" {
+			if clientID == "" {
+				clientID = d.cloud.Config.AzureAuthConfig.UserAssignedIdentityID
+			}
+			sensitiveMountOptions = []string{"sec=krb5,cruid=0,upcall_target=mount", fmt.Sprintf("username=%s", clientID)}
+			klog.V(2).Infof("using managed identity %s for volume %s with mount options: %v", clientID, volumeID, sensitiveMountOptions)
```
mittachaitu, line 413: Q: Is it good idea to log sensitiveMountOptions which has clientID?

andyzhangx: these mount options `sec=krb5,cruid=0,upcall_target=mount` including `clientID` is actually not sensitiveMountOptions, it's yes for using account key auth (not in this case)

charts/latest/azurefile-csi-driver/templates/csi-azurefile-node.yaml
andyzhangx: yes, that could make sure changes in this path could be reflected in other containers, we set as Bidirectional for this path in all containers in this driver.

mittachaitu: I agree that **mount** changes made in this will be visible to other containers, but we are not performing any mount related operations in init container so `hostToContainer` propagation should be good enough

PR #2687, opened 2025-07-23
fix: intree volume unmount failure

andyzhangx:
**What type of PR is this?**
/kind bug

fix: intree volume unmount failure

though intree volume is not supported any more, we still want to address the unmount failure if there is solution with best efforts

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

<details>

```
NodeUnpublishVolume: remove direct volume mount info #storage--storage-test-rotate#arc-appint-stage-p-env1northcentralusstagerotate4830#volume1#containerapps-ns1 from /var/lib/kubelet/pods/5c80b21a-52ee-42c9-bd96-3c66bbd2a108/volumes/kubernetes.io~csi/#xx#xx#volume1#containerapps-ns1/mount azurefile E0723 03:20:44.829002       1 utils.go:110] GRPC error: rpc error: code = Internal desc = failed to direct volume remove mount info /var/lib/kubelet/pods/5c80b21a-52ee-42c9-bd96-3c66bbd2a108/volumes/kubernetes.io~csi/#xx#xx#volume1#containerapps-ns1/mount: unlinkat /run/kata-containers/shared/direct-volumes/xx-xx==: file name too long  
```

</details>

**Release note**:
```
fix: intree volume unmount failure
```

### DIFF
--- a/pkg/azurefile/nodeserver.go
+++ b/pkg/azurefile/nodeserver.go
@@ -219,6 +219,10 @@ func (d *Driver) NodeUnpublishVolume(_ context.Context, req *csi.NodeUnpublishVo
 		// Remove deletes the direct volume path including all the files inside it.
 		// if there is no kata-cc mountinfo present on this path, it will return nil.
 		if err := d.directVolume.Remove(targetPath); err != nil {
+			if strings.Contains(err.Error(), "file name too long") {
+				klog.Warningf("NodeUnpublishVolume: direct volume mount info %s not found on %s, ignoring error", volumeID, targetPath)
+				return &csi.NodeUnpublishVolumeResponse{}, nil
+			}
 			return nil, status.Errorf(codes.Internal, "failed to direct volume remove mount info %s: %v", targetPath, err)
 		}
 	}


nearora-msft:
@andyzhangx is this issue applicable only to the in-tree volumes? 
I don't think it should matter whether it is in-tree or csi?

andyzhangx:
> @andyzhangx is this issue applicable only to the in-tree volumes? I don't think it should matter whether it is in-tree or csi?

@nearora-msft in-tree volume would make the mount path longer than csi volume, and from what users have reported, all such issues are due to using in-tree volume, I don't see there is csi volume hitting such issue up till now, do you have any?

nearora-msft:
> > @andyzhangx is this issue applicable only to the in-tree volumes? I don't think it should matter whether it is in-tree or csi?
> 
> @nearora-msft in-tree volume would make the mount path longer than csi volume, and from what users have reported, all such issues are due to using in-tree volume, I don't see there is csi volume hitting such issue up till now, do you have any?

Oh I see, thanks for responding. No I haven't seen any other than this.

nearora-msft:
> > > @andyzhangx is this issue applicable only to the in-tree volumes? I don't think it should matter whether it is in-tree or csi?
> > 
> > 
> > @nearora-msft in-tree volume would make the mount path longer than csi volume, and from what users have reported, all such issues are due to using in-tree volume, I don't see there is csi volume hitting such issue up till now, do you have any?
> 
> Oh I see, thanks for responding. No I haven't seen any other than this.

@andyzhangx on another note, why was enableKataCCMount path invoked for the customer? Shouldn't it be disabled by default unless the cx has set it in storageclass or PV?

andyzhangx:
> > > > @andyzhangx is this issue applicable only to the in-tree volumes? I don't think it should matter whether it is in-tree or csi?
> > > 
> > > 
> > > @nearora-msft in-tree volume would make the mount path longer than csi volume, and from what users have reported, all such issues are due to using in-tree volume, I don't see there is csi volume hitting such issue up till now, do you have any?
> > 
> > 
> > Oh I see, thanks for responding. No I haven't seen any other than this.
> 
> @andyzhangx on another note, why was enableKataCCMount path invoked for the customer? Shouldn't it be disabled by default unless the cx has set it in storageclass or PV?

@nearora-msft that would require an extra config in sc if user wants to mount pv on kata enabled node, with current design, as long as it's kata enabled node, user could mount azure file volume directly without any extra config.

Issue #1296, opened 2023-06-08
volumeAttributes.folderName fails of folder not exists on share


xrow:
**What happened**:

I have used the guide to come with your own storage account

My PersistentVolume has the option to moun to a subfolder. The folder doesn`t not yet exist on the share.

```yaml
    volumeAttributes:
      folderName: files
```

Mount will not attach.

**What you expected to happen**:

Mount will attach.

**How to reproduce it**:

```yaml
    volumeAttributes:
      folderName: notexistingfolderonshare
```

**Anything else we need to know?**:

If you create the subfolder manually on the share via azure storage explorer it will attach.

**Environment**:

- CSI Driver version: mcr.microsoft.com/oss/kubernetes-csi/azurefile-csi:v1.26.3
- Kubernetes version (use `kubectl version`): latest aks
- OS (e.g. from /etc/os-release): latest aks
- Kernel (e.g. `uname -a`): latest aks
- Install tools: latest aks
- Others: latest aks

andyzhangx:
that's expected per the parameter doc: https://github.com/kubernetes-sigs/azurefile-csi-driver/blob/master/docs/driver-parameters.md

folderName | specify folder name in Azure file share | existing folder name in Azure file share | No | if folder name does not exist in file share, mount would fail

xrow:
@andyzhangx True sorry... In that case I would like to request this as a feature. So create that folder if it doen`t exist.

david-garcia-garcia:
/remove-lifecycle stale

I also came a across the need for this :(

PR #2699, opened 2025-08-01
New folderNameAutocreate

david-garcia-garcia:
**What type of PR is this?**

/kind feature

Adds a new volume attribute "folderNameAutocreate" to handle automatic creation of nested folders in the target files shares if they do not exist.

Fixes https://github.com/kubernetes-sigs/azurefile-csi-driver/issues/1296

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/azurefile/azurefile.go
+++ b/pkg/azurefile/azurefile.go
@@ -47,7 +47,6 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
-	"k8s.io/client-go/kubernetes"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/klog/v2"
 	mount "k8s.io/mount-utils"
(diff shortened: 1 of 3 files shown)


david-garcia-garcia:
Just wanted to see how difficult this was to achieve. If there is any remote chance this can make it through, I can keep working on it.

pkg/azurefile/azurefile.go
```diff
@@ -1373,3 +1373,91 @@ func isKataNode(ctx context.Context, nodeID, confidentialContainerLabel string,
...
+	// Check if the error is something other than "not found"
+	var respErr *azcore.ResponseError
+	if !errors.As(err, &respErr) || respErr.StatusCode != 404 {
+		// Some other error occurred (permissions, network, etc.)
+		return fmt.Errorf("failed to check if folder path %s exists: %w", folderName, err)
+	}
+
+	// Directory doesn't exist - need to create it recursively
+	// Split the path by '/' and create directories level by level
+	pathComponents := strings.Split(strings.Trim(folderName, "/"), "/")
```
andyzhangx, line 1406: is it necessary to support subfolder creation? from the original issue, root folder creation is enough, that could make the logic clear

david-garcia-garcia: It would make the logic way more simple, but the use cases will be ver limited. Even though the original issue only talks about first level directories - at least for me - I have several directories depth almost everywhere. Unfortunately, the API forces you to iterate level by level, I ported this implementation for a terraform exec I am using to compensate for the lack of support for this in the driver.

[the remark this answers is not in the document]
david-garcia-garcia: This is OK as it is, folderName and other volume mounting related parameters in K8S only work wih forward slashes.

andyzhangx (approved):
/lgtm

PR #2819, opened 2025-10-04
doc: mount with managed identity

andyzhangx:
**What type of PR is this?**
/kind documentation

doc: mount with managed identity

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/docs/managed-identity-mount.md
+++ b/docs/managed-identity-mount.md
@@ -5,7 +5,9 @@
 This article demonstrates the process of mounting smb file share with user-assigned managed identity authentication only, without relying on account key authentication.
 > by default you could leverage the built-in user assigned managed identity(kubelet identity) bound to the AKS agent node pool(with naming rule [`AKS Cluster Name-agentpool`](https://docs.microsoft.com/en-us/azure/aks/use-managed-identity#summary-of-managed-identities))
 
-> if you have created your own managed identity, make sure the managed identity is bound to the agent node pool.
+> if you have created your own managed identity, make sure the managed identity is associated with the agent node pool. You could use following command to bind the managed identity to the VMSS node pool:
+
+    > `az vmss identity assign --name <vmss-name> --resource-group <resource-group-name> --identities <managed-identity-resource-id>`
 
 ## Before you begin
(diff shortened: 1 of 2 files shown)


docs/managed-identity-mount.md
```diff
@@ -0,0 +1,107 @@
+# Mount Azure SMB file share with managed identity
+ - Feature status: Preview
```
mittachaitu, line 2: Here it is mentioned as `Preview` feature but this feature is enabled by default in [1.34](https://github.com/kubernetes-sigs/azurefile-csi-driver/blob/57f3c7d702673f00aca57a0f441cef37c6eb3add/charts/v1.34.1/azurefile-csi-driver/values.yaml#L130) release of files, should we need to update it to false by default (or) should we change here as supported?

docs/managed-identity-mount.md
```diff
@@ -0,0 +1,107 @@
...
+ - supported from CSI driver v1.34.0 on Linux node
+
+This article demonstrates the process of mounting smb file share with user-assigned managed identity authentication only, without relying on account key authentication.
+> by default you could leverage the built-in user assigned managed identity(kubelet identity) bound to the AKS agent node pool(with naming rule [`AKS Cluster Name-agentpool`](https://docs.microsoft.com/en-us/azure/aks/use-managed-identity#summary-of-managed-identities))
+
+> if you have created your own managed identity, make sure the managed identity is bound to the agent node pool.
```
mittachaitu: IMO, it is good to mention the az command to bind the managed identity to agent node pool
```sh
az vmss identity assign --resource-group <MC_ResourceGroupName> --name <VMSSName> --identities <UserAssignedIdentityResourceID>
```

docs/managed-identity-mount.md
andyzhangx: there is a new parameter(`mountWithManagedIdentity`) to decide whether mount with managed identity or not, that means if user does not use that parameter, it won't be enabled (no behavior change by default), so that's ok

docs/managed-identity-mount.md
andyzhangx: added

PR #2852, opened 2025-11-14
fix: snapshot restore and volume clone failure in some new sovereign clouds

hasethuraman:
**What type of PR is this?**

/kind bug

We use azcopy during volume clone. Azcopy trusts known Microsoft's endpoint suffixes. When a new endpoint is added, AzCopy fails saying it is not trustable & so volume provisioning fails. Since Azure File CSI Driver knows the trusted new endpoint through the config & in such case if the new endpoint is not in azcopy known list, we will use the option that azcopy provides to pass new suffixes

Without this PR, we need to release new version until which the volume clone is not possible

Fixes #

**Requirements**:
  
- [x] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/azurefile/controllerserver.go
+++ b/pkg/azurefile/controllerserver.go
@@ -1263,13 +1263,7 @@ func (d *Driver) execAzcopyCopy(srcPath, dstPath string, azcopyCopyOptions, auth
 
 	// Use --trusted-microsoft-suffixes option to avoid failure caused by
 	if d.requiredAzCopyToTrust {
-		cmd := exec.Command("azcopy", "copy", "--trusted-microsoft-suffixes", d.getStorageEndPointSuffix(), srcPath, dstPath)
-		if len(authAzcopyEnv) > 0 {
-			cmd.Env = append(os.Environ(), authAzcopyEnv...)
-		}
-		if out, err := cmd.CombinedOutput(); err != nil {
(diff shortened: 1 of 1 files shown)


pkg/azurefile/controllerserver.go
```diff
@@ -1257,6 +1260,18 @@ func (d *Driver) copyFileShareByAzcopy(ctx context.Context, srcFileShareName, ds
...
+	if d.requiredAzCopyToTrust {
+		cmd := exec.Command("azcopy", "copy", "--trusted-microsoft-suffixes", d.getStorageEndPointSuffix(), srcPath, dstPath)
+		if len(authAzcopyEnv) > 0 {
+			cmd.Env = append(os.Environ(), authAzcopyEnv...)
+		}
+		if out, err := cmd.CombinedOutput(); err != nil {
+			return out, fmt.Errorf("exec set-trusted-microsoft-suffixes error: %v, output: %s", err, string(out))
+		}
+	}
+
```
ganesan23, line 1268: Why do we continue to regular azcopy if the azcopy copy --trusted-microsoft-suffixes succeed? Isn't the copy already successfully done at this point?

andyzhangx (commented):
pls provide more details about this PR change in the PR description, and set as `Ready for Review` mode, so more tests would run to avoid any regression.

hasethuraman:
Thanks @andyzhangx. Updated the PR description

andyzhangx (approved):
/lgtm

PR #2861, opened 2025-11-15
[release-1.32] cherry-pick: fix: snapshot restore and volume clone failure in some new sovereign clouds

rnhan:
**What type of PR is this?**

/kind bug

Cherry pick the fix from https://github.com/kubernetes-sigs/azurefile-csi-driver/pull/2852 into **release-1.32:**

> We use azcopy during volume clone. Azcopy trusts known Microsoft's endpoint suffixes. When a new endpoint is added, AzCopy fails saying it is not trustable & so volume provisioning fails. Since Azure File CSI Driver knows the trusted new endpoint through the config & in such case if the new endpoint is not in azcopy known list, we will use the option that azcopy provides to pass new suffixes

> Without this PR, we need to release new version until which the volume clone is not possible

Fixes #

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/azurefile/controllerserver.go
+++ b/pkg/azurefile/controllerserver.go
@@ -1219,13 +1219,7 @@ func (d *Driver) execAzcopyCopy(srcPath, dstPath string, azcopyCopyOptions, auth
 
 	// Use --trusted-microsoft-suffixes option to avoid failure caused by
 	if d.requiredAzCopyToTrust {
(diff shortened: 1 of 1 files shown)


andyzhangx (approved):
/ok-to-test
/lgtm

PR #2937, opened 2026-01-12
unit tests: chmod setgid bit and allow skipping cloud tests

dobsonj:
**What type of PR is this?**

/kind test

This PR includes two small commits to improve unit tests:

- **test: add option to skip unit tests that block on cloud provider**

Some unit tests exercise back-end cloud provider functionality and take >1min to run per test, which can result in hitting the timeout. https://github.com/kubernetes-sigs/azurefile-csi-driver/commit/a7a17ca9bbf46f33e0b5fc3a3f42f56f5123b50f already bumped the timeout from 10m to 30m to help with this. However, we generally limit cloud provider tests to e2e, not unit tests. Ideally, it should be possible to run unit tests against the code without depending on availability of external services. This commit allows developers to optionally skip the long running tests with `make unit-test UNIT_TEST_ARGS=-short`

- **test: always chmod expected setgid bit in TestChmodIfPermissionMismatch**

In one environment, TestChmodIfPermissionMismatch was failing 3 cases that expected setgid NOT to be set:

```
=== RUN   TestChmodIfPermissionMismatch
    utils_test.go:525: test[permission matching path]: unexpected gid bit: true, expected gid bit: false
    utils_test.go:525: test[permission mismatch path]: unexpected gid bit: true, expected gid bit: false
    utils_test.go:525: test[only change the permission mode bits when gid is not set but mode bits have gid set]: unexpected gid bit: true, expected gid bit: false
--- FAIL: TestChmodIfPermissionMismatch (0.00s)
```

The setgid bit was never specified and chmod was never called for these cases, so the test makes an assumption about the environment where it runs. This commit calls chmod to explicitly clear setgid for those test cases to avoid making assumptions about the default behavior.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

/cc @andyzhangx

**Release note**:
```
none
```

### DIFF
--- a/pkg/azurefile/utils_test.go
+++ b/pkg/azurefile/utils_test.go
@@ -435,10 +435,12 @@ func TestChmodIfPermissionMismatch(t *testing.T) {
 	skipIfTestingOnWindows(t)
 	permissionMatchingPath, _ := getWorkDirPath("permissionMatchingPath")
 	_ = makeDir(permissionMatchingPath, 0755)
+	_ = os.Chmod(permissionMatchingPath, 0755&^os.ModeSetgid) // clear setgid bit
 	defer os.RemoveAll(permissionMatchingPath)
 
 	permissionMismatchPath, _ := getWorkDirPath("permissionMismatchPath")
 	_ = makeDir(permissionMismatchPath, 0721)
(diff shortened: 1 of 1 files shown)


andyzhangx (approved):
/lgtm

PR #3148, opened 2026-05-11
deprecate allow-inline-volume-key-access-with-identity flag (no-op)

srranjan-sri:
**What type of PR is this?**

This PR adds a warning to inform users that the allow-inline-volume-key-access-with-identity flag will be deprecated in a future release when it is set to true.
image

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


mittachaitu (approved):
LGTM!!

pkg/azurefile/azurefile.go
```diff
@@ -337,7 +336,9 @@ func NewDriver(options *DriverOptions) *Driver {
 	driver.customUserAgent = options.CustomUserAgent
 	driver.userAgentSuffix = options.UserAgentSuffix
 	driver.allowEmptyCloudConfig = options.AllowEmptyCloudConfig
-	driver.allowInlineVolumeKeyAccessWithIdentity = options.AllowInlineVolumeKeyAccessWithIdentity
```
andyzhangx, line 345: it needs to be `driver.allowInlineVolumeKeyAccessWithIdentity = options.AllowInlineVolumeKeyAccessWithIdentity`, otherwise `options.AllowInlineVolumeKeyAccessWithIdentity` won't take effect any more

srranjan-sri: allow-inline-volume-key-access-with-identity will fall under same bucket as https://github.com/kubernetes-sigs/azurefile-csi-driver/pull/3128  and it is going to be a security concern k8s cluster so this PR disables it and making it no-affect. The line `"driver.allowInlineVolumeKeyAccessWithIdentity = options.AllowInlineVolumeKeyAccessWithIdentity"` only stores the flag value passed by the user for later use. However, since this value is not used anywhere downstream, I don’t think we need to retain this assignment.
The options.AllowInlineVolumeKeyAccessWithIdentity value is already populated automatically by Go’s flag parser from the CLI argument, and we only need to check it once during startup to emit the deprecation warning. Am I thinking about this correctly?

andyzhangx: No, that's different scenario, this `allowInlineVolumeKeyAccessWithIdentity` is for mount with account key (using managed identity to retrieve account key), we set this flag as false by default due to security concern long time ago, and provides the user the option to enable this feature. we could keep this flag as it is now otherwise it may break some upstream users.

srranjan-sri: okay, got it. reverted back this change.

pkg/azurefile/nodeserver.go
```diff
@@ -103,8 +103,8 @@ func (d *Driver) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolu
 				return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("mountWithManagedIdentity cannot be used for ephemeral volumes, please use either %s or secret based authentication", mountWithWITokenField))
 			}
 			useWIToken := strings.EqualFold(getValueInMap(context, mountWithWITokenField), trueValue)
-			if !d.allowInlineVolumeKeyAccessWithIdentity && !useWIToken {
-				// only get storage account from secret when not using managed identity or workload identity
+			if !useWIToken {
```
andyzhangx: this still changes the behavior

srranjan-sri: change reverted

charts/latest/azurefile-csi-driver/templates/csi-azurefile-node-windows.yaml
```diff
@@ -134,7 +134,7 @@ spec:
             - "--user-agent-suffix={{ .Values.driver.userAgentSuffix }}"
             - "--allow-empty-cloud-config={{ .Values.node.allowEmptyCloudConfig }}"
             - "--enable-get-volume-stats={{ .Values.feature.enableGetVolumeStats }}"
-            - "--allow-inline-volume-key-access-with-identity={{ .Values.node.allowInlineVolumeKeyAccessWithIdentity }}"
+            - "--allow-inline-volume-key-access-with-identity={{ .Values.node.allowInlineVolumeKeyAccessWithIdentity }}" # DEPRECATED: no-op, will be removed in a future release
```
landreasyan: Let's remove the no-op since the functionality stays for the next 2 or so releases.

srranjan-sri: Replaced the Deprecated: no-op comment with will be removed in a future release. The only remaining change in this PR is logging a warning when the allow-inline-volume-... flag is set to true.

mittachaitu (commented):
In template let's add a check and print that this option will be depricated in future release, so that this will get notified to user as part of upgrades/new installations

{{- if values.node.allowInlineVolumeKeyAccessWithIdentity }}
{{- printf "WARNING: 'allowInlineVolumeKeyAccessWithIdentity' will be deprecated in future releases." | quote | indent 0 }}
{{- end }}

andyzhangx (commented):
we should leave the `allow-inline-volume-key-access-with-identity` flag as it is now since only open source **admin** user can enable this feature, there is no impact on AKS managed csi driver.

The original inline volume with managed identity auth mount is already non supported.

mittachaitu:
> we should leave the `allow-inline-volume-key-access-with-identity` flag as it is now since only open source **admin** user can enable this feature, there is no impact on AKS managed csi driver.
> 
> The original inline volume with managed identity auth mount is already non supported.

Here the issue is little different even **admin** enables this flag they will not have control on pods which uses inline volume. If admin enables this flag and some arbitrary pod shouldn't access fileshare (without right set of permissions/authorization). So, irrespective of flag enablement this PR does right thing from security point for OpenSource users.

andyzhangx:
> > we should leave the `allow-inline-volume-key-access-with-identity` flag as it is now since only open source **admin** user can enable this feature, there is no impact on AKS managed csi driver.
> > The original inline volume with managed identity auth mount is already non supported.
> 
> Here the issue is little different even **admin** enables this flag they will not have control on pods which uses inline volume. If admin enables this flag and some arbitrary pod shouldn't access fileshare (without right set of permissions/authorization). So, irrespective of flag enablement this PR does right thing from security point for OpenSource users.

@srranjan-sri we could add some notes about the usage of this flag, leaving this flag as disabled by default should be ok.

PR #3211, opened 2026-06-12
fix: skip unmount in ensureMountPoint to prevent stale mount cleanup race

mittachaitu:
**What type of PR is this?**
/kind bug

This PR fixes a race condition between application container restarts and periodic `NodePublishVolume` calls triggered by `requiresRepublish: true`. During republish, **ensureMountPoint** was performing an unmount when it detected a stale or broken mount If the application container restarts at the same time as this unmount, writes to the mount path land on the container's local filesystem instead of the persistent fileshare, causing silent data loss.

Per the Kubernetes CSI documentation, when `requiresRepublish: true` is set, subsequent NodePublishVolume calls should only update the contents of the volume — not remount or alter mount state. This change removes the unmount from ensureMountPoint and instead reports the mount as already present, deferring cleanup to _NodeUnpublishVolume_.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

- Ran following script while performing test
```sh
 while true; do date; mount | grep nfs; ls -lrthas /var/lib/kubelet/pods/f5de012a-3994-4566-b78d-5104175afae8/volumes/kubernetes.io~csi/pvc-7d52523f-6b91-4753-b16e-d36e1809b3a5/mount; sleep 2; done
```
It produced following output:
```
Fri Jun 12 13:51:22 UTC 2026
f88010bbf2bfc461283e201.file.core.windows.net:/f88010bbf2bfc461283e201/pvcn-7d52523f-6b91-4753-b16e-d36e1809b3a5 on /var/lib/kubelet/plugins/kubernetes.io/csi/file.csi.azure.com/b7904ac2a05379a4e8c5e968c63e64945f578aacb48eb20f0c409ea874563159/globalmount type nfs4 (rw,relatime,vers=4.1,rsize=1048576,wsize=1048576,namlen=255,acregmin=30,acregmax=30,acdirmax=30,hard,noresvport,proto=tcp,nconnect=4,timeo=600,retrans=2,sec=sys,clientaddr=10.224.0.33,local_lock=none,addr=20.150.82.72)
f88010bbf2bfc461283e201.file.core.windows.net:/f88010bbf2bfc461283e201/pvcn-7d52523f-6b91-4753-b16e-d36e1809b3a5 on /var/lib/kubelet/plugins/kubernetes.io/csi/file.csi.azure.com/b7904ac2a05379a4e8c5e968c63e64945f578aacb48eb20f0c409ea874563159/globalmount type nfs4 (rw,relatime,vers=4.1,rsize=1048576,wsize=1048576,namlen=255,acregmin=30,acregmax=30,acdirmax=30,hard,noresvport,proto=tcp,nconnect=4,timeo=600,retrans=2,sec=sys,clientaddr=10.224.0.33,local_lock=none,addr=20.150.82.72)
f88010bbf2bfc461283e201.file.core.windows.net:/f88010bbf2bfc461283e201/pvcn-7d52523f-6b91-4753-b16e-d36e1809b3a5 on /var/lib/kubelet/pods/f5de012a-3994-4566-b78d-5104175afae8/volumes/kubernetes.io~csi/pvc-7d52523f-6b91-4753-b16e-d36e1809b3a5/mount type nfs4 (rw,relatime,vers=4.1,rsize=1048576,wsize=1048576,namlen=255,acregmin=30,acregmax=30,acdirmax=30,hard,noresvport,proto=tcp,nconnect=4,timeo=600,retrans=2,sec=sys,clientaddr=10.224.0.33,local_lock=none,addr=20.150.82.72)
f88010bbf2bfc461283e201.file.core.windows.net:/f88010bbf2bfc461283e201/pvcn-7d52523f-6b91-4753-b16e-d36e1809b3a5 on /var/lib/kubelet/pods/f5de012a-3994-4566-b78d-5104175afae8/volumes/kubernetes.io~csi/pvc-7d52523f-6b91-4753-b16e-d36e1809b3a5/mount type nfs4 (rw,relatime,vers=4.1,rsize=1048576,wsize=1048576,namlen=255,acregmin=30,acregmax=30,acdirmax=30,hard,noresvport,proto=tcp,nconnect=4,timeo=600,retrans=2,sec=sys,clientaddr=10.224.0.33,local_lock=none,addr=20.150.82.72)
total 7.5K
4.0K drwxr-x--- 3 root root  4.0K Jun 12 13:49 ..
 512 drwxrwsrwx 2 root 65532   64 Jun 12 13:49 .
3.0K -rw-r--r-- 1 root 65532 2.8K Jun 12 13:51 outfile
```
- when fileshare is deleted and unmount was performed by csi driver then application will write to local filesystem
```
Fri Jun 12 13:53:01 UTC 2026
f88010bbf2bfc461283e201.file.core.windows.net:/f88010bbf2bfc461283e201/pvcn-7d52523f-6b91-4753-b16e-d36e1809b3a5 on /var/lib/kubelet/plugins/kubernetes.io/csi/file.csi.azure.com/b7904ac2a05379a4e8c5e968c63e64945f578aacb48eb20f0c409ea874563159/globalmount type nfs4 (rw,relatime,vers=4.1,rsize=1048576,wsize=1048576,namlen=255,acregmin=30,acregmax=30,acdirmax=30,hard,noresvport,proto=tcp,nconnect=4,timeo=600,retrans=2,sec=sys,clientaddr=10.224.0.33,local_lock=none,addr=20.150.82.72)
f88010bbf2bfc461283e201.file.core.windows.net:/f88010bbf2bfc461283e201/pvcn-7d52523f-6b91-4753-b16e-d36e1809b3a5 on /var/lib/kubelet/plugins/kubernetes.io/csi/file.csi.azure.com/b7904ac2a05379a4e8c5e968c63e64945f578aacb48eb20f0c409ea874563159/globalmount type nfs4 (rw,relatime,vers=4.1,rsize=1048576,wsize=1048576,namlen=255,acregmin=30,acregmax=30,acdirmax=30,hard,noresvport,proto=tcp,nconnect=4,timeo=600,retrans=2,sec=sys,clientaddr=10.224.0.33,local_lock=none,addr=20.150.82.72)
total 12K
4.0K drwxr-xr-x 2 root root 4.0K Jun 12 13:52 .
4.0K drwxr-xr-x 3 root root 4.0K Jun 12 13:52 ..
4.0K -rw-r--r-- 1 root root  957 Jun 12 13:53 outfile
```
- State of pod directory points to ext3/ext2
```
root@aks-agentpool0-71348700-vmss000001:/# stat /var/lib/kubelet/pods/f5de012a-3994-4566-b78d-5104175afae8/volumes/kubernetes.io~csi/pvc-7d52523f-6b91-4753-b16e-d36e1809b3a5/mount -f
  File: "/var/lib/kubelet/pods/f5de012a-3994-4566-b78d-5104175afae8/volumes/kubernetes.io~csi/pvc-7d52523f-6b91-4753-b16e-d36e1809b3a5/mount"
    ID: 5aa59d5991d66bd3 Namelen: 255     Type: ext2/ext3
Block size: 4096       Fundamental block size: 4096
Blocks: Total: 32229372   Free: 25336968   Available: 25332872
Inodes: Total: 16646144   Free: 16347425
```

**Release note**:
```
none
```

### DIFF
--- a/pkg/azurefile/nodeserver.go
+++ b/pkg/azurefile/nodeserver.go
@@ -210,7 +210,7 @@ func (d *Driver) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolu
 		klog.V(2).Infof("NodePublishVolume: refreshed OAuth token credential cache for volume(%s) server(%s)", volumeID, oauthServer)
 	}
 
-	// NodePublishVolume should only alter the content of volume post initial successfull mount
+	// NodePublishVolume should only alter volume content after the initial successful mount.
 	mnt, err := d.ensureMountPoint(target, os.FileMode(mountPermissions), false)
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, "Could not mount target %s: %v", target, err)
--- a/pkg/azurefile/nodeserver.go
+++ b/pkg/azurefile/nodeserver.go
@@ -880,7 +880,7 @@ func (d *Driver) ensureMountPoint(target string, perm os.FileMode) (bool, error)
 		// The actual cleanup is deferred to NodeUnpublishVolume, which is the proper
 		// teardown path.
 		klog.Warningf("ReadDir %s failed with %v, skip unmount to avoid race condition", target, err)
-		return !notMnt, nil
+		return !notMnt, err
 	}
 	if err := makeDir(target, perm); err != nil {
 		klog.Errorf("MakeDir failed on target: %s (%v)", target, err)
--- a/pkg/azurefile/nodeserver.go
+++ b/pkg/azurefile/nodeserver.go
@@ -869,14 +869,18 @@ func (d *Driver) ensureMountPoint(target string, perm os.FileMode) (bool, error)
 			klog.V(2).Infof("already mounted to target %s", target)
 			return !notMnt, nil
 		}
-		// mount link is invalid, now unmount and remount later
-		klog.Warningf("ReadDir %s failed with %v, unmount this directory", target, err)
-		if err := d.mounter.Unmount(target); err != nil {
-			klog.Errorf("Unmount directory %s failed with %v", target, err)
-			return !notMnt, err
-		}
-		notMnt = true
-		return !notMnt, err
+		// Do not unmount here even if the mount is stale or broken (e.g., ESTALE on NFS).
+		// Unmounting during a periodic NodePublishVolume (requiresRepublish) creates a race
+		// condition: if the application container restarts at the same time, it may bind to
+		// the unmounted target path and write to the container's local filesystem instead of
+		// the persistent volume, causing silent data loss.
+		//
+		// Per the Kubernetes CSI spec, subsequent NodePublishVolume calls (republish) should
+		// only update the contents of the volume, not remount or alter mount state.
+		// The actual cleanup is deferred to NodeUnpublishVolume, which is the proper
+		// teardown path.
+		klog.Warningf("ReadDir %s failed with %v, skip unmount to avoid race condition", target, err)
+		return !notMnt, nil
 	}
 	if err := makeDir(target, perm); err != nil {
 		klog.Errorf("MakeDir failed on target: %s (%v)", target, err)
--- a/pkg/azurefile/nodeserver.go
+++ b/pkg/azurefile/nodeserver.go
@@ -210,7 +210,8 @@ func (d *Driver) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolu
 		klog.V(2).Infof("NodePublishVolume: refreshed OAuth token credential cache for volume(%s) server(%s)", volumeID, oauthServer)
 	}
 
-	mnt, err := d.ensureMountPoint(target, os.FileMode(mountPermissions))
+	// NodePublishVolume should only alter the content of volume post initial successfull mount
+	mnt, err := d.ensureMountPoint(target, os.FileMode(mountPermissions), false)
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, "Could not mount target %s: %v", target, err)
 	}
@@ -528,7 +529,7 @@ func (d *Driver) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRe
 
 	klog.V(2).Infof("cifsMountPath(%v) fstype(%v) volumeID(%v) mountflags(%v) mountOptions(%v) volumeMountGroup(%s)", cifsMountPath, fsType, volumeID, mountFlags, mountOptions, volumeMountGroup)
 
-	isDirMounted, err := d.ensureMountPoint(cifsMountPath, os.FileMode(mountPermissions))
+	isDirMounted, err := d.ensureMountPoint(cifsMountPath, os.FileMode(mountPermissions), true)
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, "Could not mount target %s: %v", cifsMountPath, err)
 	}
@@ -635,7 +636,7 @@ func (d *Driver) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRe
 	}
 
 	if isDiskMount {
-		mnt, err := d.ensureMountPoint(targetPath, os.FileMode(mountPermissions))
+		mnt, err := d.ensureMountPoint(targetPath, os.FileMode(mountPermissions), true)
 		if err != nil {
 			return nil, status.Errorf(codes.Internal, "mount %s on target %s failed with %v", volumeID, targetPath, err)
 		}
@@ -821,7 +822,7 @@ func (d *Driver) NodeExpandVolume(_ context.Context, _ *csi.NodeExpandVolumeRequ
 
 // ensureMountPoint: create mount point if not exists
 // return <true, nil> if it's already a mounted point otherwise return <false, nil>
-func (d *Driver) ensureMountPoint(target string, perm os.FileMode) (bool, error) {
+func (d *Driver) ensureMountPoint(target string, perm os.FileMode, shouldUnmount bool) (bool, error) {
 	notMnt, err := d.mounter.IsLikelyNotMountPoint(target)
 	if err != nil && !os.IsNotExist(err) {
 		if IsCorruptedDir(target) {
@@ -868,6 +869,16 @@ func (d *Driver) ensureMountPoint(target string, perm os.FileMode) (bool, error)
 		if err == nil {
 			klog.V(2).Infof("already mounted to target %s", target)
 			return !notMnt, nil
+		}
+		if shouldUnmount {
+			klog.Warningf("ReadDir %s failed with %v, unmounting stale mount to fix it", target, err)
+			// mount link is invalid, now unmount and remount later
+			if err := d.mounter.Unmount(target); err != nil {
+				klog.Errorf("Unmount directory %s failed with %v", target, err)
+				return !notMnt, err
+			}
+			notMnt = true
+			return !notMnt, err
 		}
 		// Do not unmount here even if the mount is stale or broken (e.g., ESTALE on NFS).
 		// Unmounting during a periodic NodePublishVolume (requiresRepublish) creates a race
--- a/pkg/azurefile/nodeserver_test.go
+++ b/pkg/azurefile/nodeserver_test.go
@@ -1235,34 +1235,67 @@ func TestEnsureMountPoint(t *testing.T) {
 	azureFile := "./azure.go"
 
 	tests := []struct {
-		desc        string
-		target      string
-		expectedErr error
+		desc                 string
+		target               string
+		shouldUnmount        bool
+		expectedErr          error
+		expectedUnmountCount uint
 	}{
 		{
-			desc:        "[Error] Mocked by IsLikelyNotMountPoint",
-			target:      errorTarget,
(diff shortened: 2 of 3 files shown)


pkg/azurefile/nodeserver.go
```diff
@@ -869,14 +869,18 @@ func (d *Driver) ensureMountPoint(target string, perm os.FileMode) (bool, error)
...
+		// condition: if the application container restarts at the same time, it may bind to
+		// the unmounted target path and write to the container's local filesystem instead of
+		// the persistent volume, causing silent data loss.
+		//
+		// Per the Kubernetes CSI spec, subsequent NodePublishVolume calls (republish) should
+		// only update the contents of the volume, not remount or alter mount state.
+		// The actual cleanup is deferred to NodeUnpublishVolume, which is the proper
+		// teardown path.
+		klog.Warningf("ReadDir %s failed with %v, skip unmount to avoid race condition", target, err)
+		return !notMnt, nil
```
nearora-msft: How will the application recover if we aren't umounting and remounting in the NodePublishVolume?

andyzhangx (changes requested):
Thanks for looking into this race condition. The intent makes sense — avoiding unmount during republish to prevent a window where the target path is empty. However, I have some concerns:

**1. Stale mounts will never self-heal**

With this change, when `ReadDir` fails (e.g., NFS ESTALE), `ensureMountPoint` returns `!notMnt, nil` — telling the caller the mount is fine. But it's not fine — the mount is broken. The caller (`NodePublishVolume`) won't attempt a remount because it thinks everything is healthy. This means a stale NFS mount will persist until the Pod is deleted and `NodeUnpublishVolume` is called.

For workloads with `requiresRepublish`, the whole point is to recover from transient failures. This change effectively disables that recovery path for stale mounts.

**2. The race condition concern may not apply here**

The unmount + remount happens within a single `NodePublishVolume` CSI call. Kubelet won't expose the target path to the container between these operations — the bind mount into the container is a separate step that only happens after `NodePublishVolume` returns success. If there's truly a race, it would be between concurrent `NodePublishVolume` calls, which should be serialized by kubelet's volume operation lock.

**3. Alternative approaches to consider:**
- Add a lock/mutex around the unmount+remount sequence if concurrent access is the concern
- Do a remount (mount over the stale mount) instead of unmount+remount — this is atomic and avoids the empty-path window
- Use `mount -o remount` or `mount --bind` to refresh the mount without tearing it down

**4. Missing tests**

The existing test expectations should be updated to validate the new behavior (stale mount → no unmount → return success).

Could you share more details about the specific race you observed? For example, was it between `NodePublishVolume` and a container restart, or between two concurrent CSI calls? That would help evaluate the right fix.

pkg/azurefile/nodeserver.go
mittachaitu: It will result in same behaviour as it exist before when `csidriver.spec.requireRepublish` sets to false.  In OSS we are not configuring `requiresRepublish` either in [atest](https://github.com/kubernetes-sigs/azurefile-csi-driver/blob/95b469299516dbb25a0d85dcded75560cb1c91f3/charts/latest/azurefile-csi-driver/templates/csi-azurefile-driver.yaml#L11) or [1.35.3](https://github.com/kubernetes-sigs/azurefile-csi-driver/blob/95b469299516dbb25a0d85dcded75560cb1c91f3/charts/v1.35.3/azurefile-csi-driver/templates/csi-azurefile-driver.yaml#L11) versions but we are setting only in managed version of CSI driver. 

> How will the application recover if we aren't umounting and remounting.

To my knowledge application pod should be restarted to recover from this behaviour.

mittachaitu:
> 1. Stale mounts will never self-heal
>
> With this change, when ReadDir fails (e.g., NFS ESTALE), ensureMountPoint returns !notMnt, nil telling the caller the mount is fine. 

Yes thanks for pointing(somehow I missed it), with this change when ReadDir fails then I will `return !noMnt, err` (keeping behaviour as earlier).  


Updates this behaviour in second commit

mittachaitu:
> 2. The race condition concern may not apply here
>
> The unmount + remount happens within a single NodePublishVolume CSI call. Kubelet won't expose the target path to the container between these operations — the bind mount into the container is a separate step that only happens after NodePublishVolume returns success. If there's truly a race, it would be between concurrent NodePublishVolume calls, which should be serialized by kubelet's volume operation lock.

Here race is not between concurrent calls of `NodePublishVolume` or not between CSI operations. The race is happening between NodePublishVolume (part of `requiresRepublish`) performing unmount and at same time if container restarts *In-Plac*. Initial bindings are already happen in first call of NodePublishVolume of pod, so bindings already exist due to initial bindings and restart of application container restarts will directly consumes mount point which is unmounted by CSI-driver and this results application to write on local volume.

mittachaitu:
> Use mount -o remount or mount --bind to refresh the mount without tearing it down

I will evaluate behaviour with above suggestion. I hope doing `-o remount` will not perform detach of filesystem. I think this will be safer operation since kernel doesn't detach mount even application tries to write IO operations will fail.

andyzhangx:
> requiresRepublish

in that case, I think in NodePublishVolume, after `ensureMountPoint`, it would do bind mount, so you could add a new param in `ensureMountPoint` func indicating whether perform unmount or not, in NodeStageVolume, we should always perform unmount in `ensureMountPoint`, while in NodePublishVolume, skip unmount, that would return error in the end instead of writing to local disk.

mittachaitu:
> > requiresRepublish
> 
> in that case, I think in NodePublishVolume, after `ensureMountPoint`, it would do bind mount, so you could add a new param in `ensureMountPoint` func indicating whether perform unmount or not, in NodeStageVolume, we should always perform unmount in `ensureMountPoint`, while in NodePublishVolume, skip unmount, that would return error in the end instead of writing to local disk.

Hmm.. much better since it is only bind mount to host mount performing unmount/re-mouny also doesn't makesense. Will make the changes

mittachaitu:
> > requiresRepublish
> 
> in that case, I think in NodePublishVolume, after `ensureMountPoint`, it would do bind mount, so you could add a new param in `ensureMountPoint` func indicating whether perform unmount or not, in NodeStageVolume, we should always perform unmount in `ensureMountPoint`, while in NodePublishVolume, skip unmount, that would return error in the end instead of writing to local disk.

@andyzhangx , I have updated the code.. PTAL

pkg/azurefile/nodeserver.go
mittachaitu: Updated the code as part of latest commit, we are not performing unmount of bind mounts but performing unmount and mount in NodeStageVolume call.

pkg/azurefile/nodeserver.go
```diff
@@ -869,13 +870,27 @@ func (d *Driver) ensureMountPoint(target string, perm os.FileMode) (bool, error)
 			klog.V(2).Infof("already mounted to target %s", target)
 			return !notMnt, nil
 		}
-		// mount link is invalid, now unmount and remount later
-		klog.Warningf("ReadDir %s failed with %v, unmount this directory", target, err)
```
andyzhangx, line 873: would you keep the warning logs here?

andyzhangx: nvm

pkg/azurefile/nodeserver.go
```diff
@@ -869,13 +870,27 @@ func (d *Driver) ensureMountPoint(target string, perm os.FileMode) (bool, error)
...
+		// condition: if the application container restarts at the same time, it may bind to
+		// the unmounted target path and write to the container's local filesystem instead of
+		// the persistent volume, causing silent data loss.
+		//
+		// Per the Kubernetes CSI spec, subsequent NodePublishVolume calls (republish) should
+		// only update the contents of the volume, not remount or alter mount state.
+		// The actual cleanup is deferred to NodeUnpublishVolume, which is the proper
+		// teardown path.
+		klog.Warningf("ReadDir %s failed with %v, skip unmount to avoid race condition", target, err)
 		return !notMnt, err
```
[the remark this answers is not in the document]
mittachaitu: > When shouldUnmount is false (used by NodePublishVolume), a failed ReadDir returns a non-nil error. That makes NodePublishVolume fail every time for a stale/broken mount without attempting cleanup

This should be expected behavior, since underlying mount is not in good state... application needs to be re-schedule to same node or different node to mitigate this issue.

pkg/azurefile/nodeserver_test.go
```diff
@@ -1235,34 +1235,67 @@ func TestEnsureMountPoint(t *testing.T) {
...
+			expectedErr:          &os.PathError{Op: "open", Path: "./false_is_likely_target", Err: syscall.ENOENT},
+			expectedUnmountCount: 0, // Since Unmount flag is false, unmount count should be 0
+		},
+		{
+			desc:                 "[Error] Opening file with shouldUnmount false and unmount will return error shouldn't alter expected error",
+			target:               falseTarget,
+			shouldUnmount:        false,
+			expectedErr:          &os.PathError{Op: "open", Path: "./false_is_likely_target", Err: syscall.ENOENT},
+			expectedUnmountCount: 0,
+		},
```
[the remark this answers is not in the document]
mittachaitu: Yeah, Since calling unmount is not expected it should return the same error. In future if someone touches this code path and do call the unmount and it is expected to fix the test case also.

andyzhangx (approved):
/lgtm

PR #3215, opened 2026-06-15
[release-1.35] chore: add requiresRepublish to v1.35.3 helm chart

andyzhangx:
Cherry-pick of #3213 to release-1.35 branch.

Adds `requiresRepublish` field to the v1.35.3 helm chart CSIDriver template, aligning with the static `deploy/v1.35.3/csi-azurefile-driver.yaml` manifest which already has `requiresRepublish: true`.

Uses `hasKey` guard to default to `true` when the value is not present (e.g. `helm upgrade --reuse-values` from older releases).

/kind cleanup


mittachaitu (approved):
LGTM!!

PR #3231, opened 2026-06-26
fix(windows): self-heal stale node-global SMB session for subPath pods on NodePublishVolume

tdihp:
**What type of PR is this?**

/kind bug

On **Windows** node pools, a newly created pod that mounts an Azure File (SMB) volume
with `subPath` fails at startup once the node-global SMB session (created by
`New-SmbGlobalMapping` during `NodeStageVolume`) has gone stale or been torn down
(credential rotation, idle/NAT teardown, storage maintenance, etc.):

```
MountVolume.SetUp failed ... failed to prepare subPath for volumeMount "..." : Access is denied.
```

`NodeStageVolume` runs only once per volume per node, so its existing self-heal
(`IsSmbMapped` → `PathValid` → `RemoveSmbGlobalMapping` → `New-SmbGlobalMapping`) never
re-runs for subsequently created pods. Those new pods re-traverse the share (e.g. kubelet's
subPath preparation) against the dead session and fail until the node is replaced. Existing
pods keep working on already-open handles, which makes the failure look intermittent.

This PR invokes the same self-heal on the per-pod `NodePublishVolume` path:

- `NodePublishVolume` now probes the node-global SMB mapping backing the staged source
  (cheap, read-only) and, only when it is proven stale/missing, resolves credentials and
  remaps before the bind/symlink. Healthy publishes and co-tenant pods on the same share
  are unaffected (credentials are resolved lazily, only when a remap is actually needed).
- Revalidation is **serialized per node-global SMB mapping** (keyed by the UNC remotePath)
  so concurrent `NodePublishVolume` calls cannot race `RemoveSmbGlobalMapping` +
  `New-SmbGlobalMapping` and clobber each other's mapping.
- `PathValid` now bounds its `GetFileAttributes` probe with a timeout (a wedged UNC path can
  block indefinitely) and deduplicates concurrent probes per path via `singleflight`, so
  repeated kubelet retries against a dead session cannot accumulate blocked goroutines.
- The shared probe logic is factored into a single helper reused by both the stage and
  publish paths.

The change is **Windows-only** in effect (`readStagedRemotePath` returns empty on
Linux/Darwin, making the new path a no-op there).

Fixes #

**Requirements**:
- [ ] includes documentation
- [x] adds unit tests

- New host-process mounter unit tests cover the remap, remap-error, credential-error and
  empty-path cases. Windows-tagged code is compile-checked via `GOOS=windows go vet`.
- A new Windows e2e spec (`should self-heal a stale node-global SMB session for newly
  created subPath pods`) reproduces the failure: it invalidates the node-global mapping
  with a hostprocess `Remove-SmbGlobalMapping`, then deploys a fresh subPath pod and asserts
  it self-heals and reaches Running.
- Validated end-to-end on a Windows Server 2022 AKS node pool: the focused Windows e2e suite
  (mount-options / subpath / self-heal / deployment write-read) passes with no regression,
  and a stock build reproduces the exact customer error under the same scenario.

**Release note**:
```
Fix Azure File (SMB) subPath mount failures ("failed to prepare subPath ... Access is denied") for newly created pods on Windows nodes after the node-global SMB session goes stale, by revalidating and, if needed, remapping the SMB session on NodePublishVolume.
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -32,6 +32,7 @@ require (
 	go.uber.org/goleak v1.3.0
 	go.uber.org/mock v0.6.0
 	golang.org/x/net v0.56.0
+	golang.org/x/sync v0.21.0
 	golang.org/x/sys v0.46.0
 	google.golang.org/grpc v1.81.1
 	google.golang.org/protobuf v1.36.11
@@ -133,7 +134,6 @@ require (
 	golang.org/x/exp v0.0.0-20250911091902-df9299821621 // indirect
 	golang.org/x/mod v0.36.0 // indirect
 	golang.org/x/oauth2 v0.36.0 // indirect
-	golang.org/x/sync v0.21.0 // indirect
 	golang.org/x/term v0.44.0 // indirect
 	golang.org/x/text v0.38.0 // indirect
 	golang.org/x/time v0.15.0 // indirect
--- a/pkg/azurefile/azure_common_darwin.go
+++ b/pkg/azurefile/azure_common_darwin.go
@@ -31,6 +31,14 @@ func SMBMount(m *mount.SafeFormatAndMount, source, target, fsType string, option
 	return nil
 }
 
+func RevalidateSMBMount(_ *mount.SafeFormatAndMount, _ string, _ func() (username, password string, err error)) error {
+	return nil
+}
+
+func readStagedRemotePath(_ string) (string, error) {
+	return "", nil
+}
+
 func SMBUnmount(m *mount.SafeFormatAndMount, target string, _, _ bool) error {
 	return nil
 }
--- a/pkg/azurefile/azure_common_linux.go
+++ b/pkg/azurefile/azure_common_linux.go
@@ -33,6 +33,18 @@ func SMBMount(m *mount.SafeFormatAndMount, source, target, fsType string, option
 	return m.MountSensitive(source, target, fsType, options, sensitiveMountOptions)
 }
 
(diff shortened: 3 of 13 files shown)


andyzhangx:
Thanks for the detailed PR. I have a concern about the overall approach:

If the SMB mount connection is stale, the existing Windows app pods on that node would also hit **write errors intermittently** (not just new pods failing at subPath preparation). This PR only fixes the symptom for **newly created pods** by re-establishing the mapping at `NodePublishVolume` time, but existing pods with open handles against the stale session will still experience IO failures.

This suggests the root cause is **the SMB session itself becoming unstable** (network layer, NAT timeout, storage-side disconnection, credential expiration, etc.), and the CSI driver self-heal is a workaround that may **hide the underlying infrastructure issue** that needs to be fixed.

Before adding this complexity to the CSI driver, I think we should:

1. **Identify why the node-global SMB session goes stale** — is it idle timeout? NAT pinhole expiration? Azure Storage-side maintenance? Credential rotation without proper session refresh?
2. **Fix at the network/storage layer** — e.g., SMB keepalive tuning, NAT timeout adjustment, or ensuring credential rotation doesn't tear down existing sessions.
3. **Understand the frequency** — if this happens regularly, it points to a systemic infra issue; if it's a rare edge case, the workaround might be acceptable but should be documented as such.

Could you share more context on the environment where this was observed? Specifically:
- How frequently does the session go stale?
- What triggers it (idle time, credential rotation, network change)?
- Do existing pods on the same node also see IO errors when this happens?

/hold

PR #3274, opened 2026-07-20
fix: prevent ARM ListKeys throttling on kubelet requiresRepublish for clientID-based Azure File mounts

nearora-msft:
**What type of PR is this?**

Issue: 

When a CSI volume is mounted with requiresRepublish=true (which the Azure File driver uses for auth modes that need periodic credential refresh), kubelet re-invokes NodePublishVolume for every pod × volume pair roughly every 2 minutes for the lifetime of the pod.

For mounts authenticated with a clientID-based federated identity clientID is set, mountWithWIToken is not set), each republish caused the following waste per pod per 2 min:

NodePublishVolume re-invoked NodeStageVolume.
NodeStageVolume called GetAccountInfo, which for the clientID branch calls d.cloud.GetStorageAccesskeyFromServiceAccountToken→ ARM ListKeys.
NodeStageVolume then reached ensureMountPoint, saw the target was already mounted, and short-circuited — throwing the freshly fetched account key away.
At scale (many pods × many volumes × 30 calls/hour each) this drove ARM ListKeys throttling.

The WI-token path [mountWithWIToken=true] legitimately needs to run on every republish so setCredentialCachecan rotate the Kerberos ticket, so it must not be affected.

Fix:

In both NodePublishVolume branches that would re-invoke NodeStageVolume on republish (SA-token forward path and ephemeral path), detect the case where:

mountWithWIToken is not true, and
the target path is already a mount point (IsLikelyNotMountPoint returns (false, nil))
…and short-circuit with a successful response, skipping the redundant NodeStageVolume re-invocation entirely. WI-token flows still fall through unchanged.

A small helper Driver.canSkipRepublishNodeStage(context, target) encapsulates the condition and is used by both call sites.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


andyzhangx (commented):
LGTM, @nearora-msft could you fix the windows ut failure?

pkg/azurefile/nodeserver.go
```diff
@@ -1064,3 +1073,19 @@ func shouldUseServiceAccountToken(attrib map[string]string) bool {
...
+// canSkipRepublishNodeStage reports whether a NodePublishVolume call is a kubelet
+// requiresRepublish retry (target already mounted) whose auth mode has no time-bound
+// credential to refresh. When true, callers should return success without re-invoking
+// NodeStageVolume, avoiding wasteful ARM ListKeys calls (clientID-only mounts) or
+// kube-apiserver Secret.Get calls (secret-based ephemeral mounts) whose result would
+// be discarded because NodeStageVolume's ensureMountPoint short-circuits on an
+// existing mount. The WI-token path is excluded so setCredentialCache continues to
+// rotate the Kerberos ticket on every republish.
+func (d *Driver) canSkipRepublishNodeStage(context map[string]string, target string) bool {
+	if strings.EqualFold(getValueInMap(context, mountWithWITokenField), trueValue) {
```
mittachaitu, line 1086: Shouldn't we also return _false_ while mounting through `mountwithoauthtoken` for the completeness of function?

mittachaitu (commented):
Left small comment!!

pkg/azurefile/nodeserver.go
nearora-msft: I don't think that NodeStage is called from NodePublish when mountwithoauthtoken  is set.

mittachaitu: True, but I am suggesting for completeness of functionality

pkg/azurefile/nodeserver.go
```diff
@@ -1064,3 +1073,19 @@ func shouldUseServiceAccountToken(attrib map[string]string) bool {
...
+		return false
+	}
+	notMnt, err := d.mounter.IsLikelyNotMountPoint(target)
+	return err == nil && !notMnt
+}
```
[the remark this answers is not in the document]
nearora-msft: This shouldn't be confusing as context is used for volumeContext throughout code.

andyzhangx (approved):
/lgtm

PR #3299, opened 2026-07-30
docs: clarify useDataPlaneAPI: oauth support for private storage accounts

### LINKED ISSUE #2121: Snapshot restore does not work if the storage account has limited access.
**What happened**:
I wanted to restore data from a CSI snapshot but the PVC was stuck in "Pending" state due to the following error:
```
failed to provision volume with StorageClass "az-test-sc": rpc error: code = Unknown desc = exec error: exit status 1, output: INFO: Scanning...
INFO: Login with identity succeeded.
INFO: Authenticating to destination using Azure AD
INFO: Authenticating to source using Azure AD
INFO: Failed to create one or more destination container(s). Your transfers may still succeed if the container already exists.
INFO: Any empty folders will be processed, because source and destination both support folders
INFO: Failed to scan directory/file . Logging errors in scanning logs.

failed to perform copy command due to error: cannot start job due to error: GET https://<fileshare>.file.core.usgovcloudapi.net/<share-pvc-name>
--------------------------------------------------------------------------------
RESPONSE 403: 403 This request is not authorized to perform this operation.
ERROR CODE: AuthorizationFailure
--------------------------------------------------------------------------------
<?xml version="1.0" encoding="utf-8"?><Error><Code>AuthorizationFailure</Code><Message>This request is not authorized to perform this operation.
RequestId:xxxxxx
Time:xxxxxxx</Message></Error>
--------------------------------------------------------------------------------
.
```

The storage account mentioned above does not allow public access - it only allows access from the AKS VNet.

Since the AKS VNet was added to the list of networks that can access the storage account, I expected the restore to work.

**How to reproduce it**:
1. Create an AKS cluster with a user-assigned managed identity that has enough permissions to access the storage account.
2. Create a storage account and enable access to it only from AKS VNet (Go to "Networking" -> "Firewalls and virtual networks" -> "Enabled from selected virtual networks and IP addresses" and select cluster VNet and subnet)
3. Create a StorageClass on the cluster that points to the previously created storage account:
```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: az-test-sc
mountOptions:
- nouser_xattr
- mfsymlinks
- nosharesock
- actimeo=30
parameters:
  resourceGroup: <storage-account-resource-group>
  skuName: Standard_LRS
  storageAccount: <storage-account-name>
  server: <storage-account-name>.file.core.usgovcloudapi.net
reclaimPolicy: Retain
volumeBindingMode: Immediate
allowVolumeExpansion: true
```
4. Deploy a simple app that is set to use the previously created StorageClass.
```yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: az-test-pvc
  namespace: az-test-ns
spec:
  accessModes:
    - ReadWriteOnce
  storageClassName: az-test-sc
  resources:
    requests:
      storage: 25Gi
---
apiVersion: v1
kind: Pod
metadata:
  name: az-test-pod
  namespace: az-test-ns
spec:
  containers:
  - image: mcr.microsoft.com/oss/nginx/nginx:1.15.5-alpine
    name: test-pod
    resources:
      requests:
        cpu: 100m
        memory: 128Mi
      limits:
        cpu: 250m
        memory: 256Mi
    volumeMounts:
      - name: azure
        mountPath: /mnt/data
  volumes:
  - name: azure
    persistentVolumeClaim:
      claimName: az-test-pvc
```
5. Take a snapshot of the app using Velero or manually by creating a VolumeSnapshot resource.
6. Try to restore the app from the taken snapshot.
7. The PVC will be stuck in "Pending" with the error mentioned above.

To be sure it's not a problem with the AKS setup, I provisioned a VM and assigned the same identity the cluster was set up with. Also, I added the VM's VNet to the list of VNets that can access the storage account. Finally, I installed `azcopy` command on the VM and verified that it could access the file share.

**Environment**:

- CSI Driver version: 1.30.5
- Kubernetes version (use `kubectl version`): 1.29.5
- Kernel (e.g. `uname -a`):
- Install tools: The app was deployed on AKS cluster on Azure Government cloud
- Others:

andyzhangx:
**What type of PR is this?**
/kind documentation

The current docs tell users that snapshot / cloning / data-plane share operations "would fail when there is firewall or vnet setting on storage account", and the snapshot and cloning READMEs specifically instruct AKS users to set the storage account to `Enable from all networks`. This is no longer accurate when `useDataPlaneAPI: oauth` is used with the AKS-managed CSI driver: the OAuth token issued for the controller identity carries a trusted-service claim that Storage recognizes via the default `networkAcls.bypass=AzureServices`, so snapshot / cloning / share create/delete/resize now work against private storage accounts (Selected networks / Public disabled / Private Endpoint) as long as:

1. CSI driver v1.33.0+ is used (v1.33.4+ for NFS snapshot restore).
2. AKS release includes the trusted-service audience configuration (rolled out mid-2026).
3. Driver controller identity has `Storage File Data Privileged Contributor` on the storage account.
4. Storage account `networkAcls.bypass` includes `AzureServices` (default).

**Changes**:

- `docs/driver-parameters.md`
  - Expand `useDataPlaneAPI` description in both `StorageClass` and `VolumeSnapshotClass` tables, distinguishing the behavior of `true`, `false`, and `oauth`.
  - Add a new section `Using useDataPlaneAPI: oauth against private storage accounts` covering the full prerequisites, an example `az role assignment`, and limitations (node mount still needs a reachable network path; self-managed installs do not get the trusted-service bypass).
- `deploy/example/snapshot/README.md`
  - Replace the misleading limitation ("configure the storage account's VNet settings to Enable from all networks") with correct configuration for public / private / self-managed cases.
  - Add an example `VolumeSnapshotClass` with `useDataPlaneAPI: "oauth"`.
- `deploy/example/cloning/README.md`
  - Remove the stale note that private endpoints are not supported.
  - Same public / private / self-managed clarification and example snippet.

Fixes #2121

Docs-only change. Behavior described here has been end-to-end validated against private storage accounts by the Azure Storage team across SMB and NFS (all networks / Selected networks / Public disabled, with and without Private Endpoint).

**Release note**:

\`\`\`release-note
NONE
\`\`\`


andyzhangx:
Addressed all 5 review comments in commit 84d90bd:

1. **`useDataPlaneAPI=true` (StorageClass row, line ~72)**: rewrote to say the controller must have network reachability to the data plane endpoint; firewall/vnet rules only block when the controller's source network is not allowlisted; only `publicNetworkAccess=Disabled` is inherently blocking.
2. **Same wording (VolumeSnapshotClass row, line ~142)**: applied the same fix.
3. **Private-account section (line ~185)**: separated authorization (OAuth trusted-service claim = *who* can call the data plane) from routing (network path to the endpoint). Replaced hardcoded `file.core.windows.net` with `<storage-endpoint-suffix>` and listed the public / China / US Government suffixes explicitly.
4. **`allowSharedKeyAccess = false` (line ~197)**: qualified. The controller rejects the default `storeAccountKey=true` when shared key access is disabled, which breaks SMB flows. Recommendation now restricted to NFS-only workloads (or users who switch to managed-identity SMB mount).
5. **`deploy/example/snapshot/README.md` and `cloning/README.md`**: replaced the hardcoded `file.core.windows.net` with `<storage-endpoint-suffix>` and the same set of sovereign-cloud examples.

PTAL.

PR #3300, opened 2026-07-30
docs: expand MI mount troubleshooting with real diagnostic commands

andyzhangx:
**What type of PR is this?**
/kind documentation

**What this PR does / why we need it:**

`Error calling AzAuthenticatorLib: -1` is a generic error code that surfaces for very different root causes (IMDS token acquisition, kerberos ticket exchange, clock skew, storage firewall, missing role assignment). Today the troubleshooting section only lists two config checks, which is not enough for support engineers or users triaging real incidents.

This PR expands the section with:
- Concrete commands to run inside the `csi-azurefile-node` pod on the failing node
- IMDS token check (`curl 169.254.169.254`)
- Verbose `azfilesauthmanager` run to surface the real error
- `dmesg` / syslog checks for CIFS + kerberos upcall errors
- A quick-reference table mapping common error strings to their root cause

Based on real customer incidents on AKS 1.34.7 where the generic `-1` error was actually caused by the MI not being propagated to the VMSS instance.

**Which issue(s) this PR fixes:**
N/A

**Requirements:**
- [x] squashed commits
- [x] includes documentation

**Release note:**
```release-note
NONE
```

### DIFF
--- a/docs/managed-identity-mount.md
+++ b/docs/managed-identity-mount.md
@@ -168,3 +168,71 @@ Verify the following:
 1. **The managed identity has the correct role assignment.** Ensure the managed identity is assigned the **`Storage File Data SMB MI Admin`** role on the storage account (or the resource group for dynamic provisioning). Other roles such as `Storage File Data SMB Share Contributor` or `Storage File Data SMB Share Elevated Contributor` are **not sufficient** for managed identity mount.
 
 2. **The SMBOauth property is enabled on the storage account.** Without this, the storage account does not support Kerberos ticket acquisition for managed identity authentication.
+
+   ```bash
+   az storage account show --name <storageAccountName> --resource-group <resourceGroup> \
+     --query "smbOAuthSettings.isSmbOAuthEnabled"
+   ```
+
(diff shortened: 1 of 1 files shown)


andyzhangx:
Addressed all 5 review comments in commit 76dc643:

1. `jq` → `--query 'userAssignedIdentities'` (portable, no extra tool needed).
2. IMDS check now queries the driver's actual OAuth resource `https://storage.azure.com/` and pipes through Python to **redact the `access_token`** in the output, with a warning comment that it's a bearer credential.
3. Explicitly note that `az vmss identity show` returns the VMSS *model* only, and IMDS from the node is the authoritative per-instance propagation check.
4. Dropped hard-coded `/usr/bin/azfilesauthmanager` in favor of the bare command via `PATH`.
5. Rewrote the `KRB5KDC_ERR_S_PRINCIPAL_UNKNOWN` row: focus on storage account name/endpoint correctness and SMB OAuth enablement instead of SMB Multichannel/3.1.1 (which is client-negotiated and not user-toggleable in this scenario).

PTAL.

Issue #2277, opened 2024-12-16
unnecessary storage service endpoint created on subnet?


kewalaka:
**What happened**:

We pre-create a storage account with a private endpoint.

When creating an NFS share via a storage class & PVC, a service endpoint is created on the subnet.

**What you expected to happen**:

It's unclear to me why storage accounts that are accessed via private endpoint need a service endpoint provisioned?

The additional service endpoint then requires routes on the UDR if we wish to avoid situations where operators provision additional storage without private endpoints and traffic bypasses the next-hop firewall because the service endpoint routing is more specific.

Ideally, we want a policy in place to block service endpoints.

**Anything else we need to know?**:

Can workaround this by adding a route, but preference is to avoid unnecessary service endpoints.

andyzhangx:
are your node pools on that subnet? @kewalaka

kewalaka:
hi @andyzhangx - yes, the service endpoint has been made at the vnet and applies to all subnets in that vnet.

in my situation I have three subnets

1) nodes pools
2) api server
3) a base subnet where we place some private endpoints

andyzhangx:
@kewalaka could you set `subnetName: nodepoolSubnetName`  in storage class, then only that subnet would be created with service endpoint, thx.

kewalaka:
true - does the node pool even need it though, if the target storage account has a private endpoint?  

That's what I was meaning, the service endpoint seems redundant.  

Or does the node pool require this for internal reasons?

andyzhangx:
@kewalaka 
what is your storage class config? if `networkEndpointType: privateEndpoint` is not set, the driver would update the subnet with service endpoint.

kewalaka:
unless i'm mistaken, this seems to be creating the private endpoint ?  this is already in place, so is undesirable.

andyzhangx:
from the driver code, the driver would only update subnet (add Microsoft.Storage service endpoint) for nfs volume only when `networkEndpointType: privateEndpoint` is not set in storage class:

https://github.com/kubernetes-sigs/azurefile-csi-driver/blob/a277f7a3ee13a7dca8cb30e424402fbfe681bdfa/pkg/azurefile/controllerserver.go#L356-L362

andyzhangx:
so if you want to avoid creating unnecessary storage service endpoint on the storage account you have provided, pls set `networkEndpointType: privateEndpoint` in your storage class

PR #3304, opened 2026-07-31
feat: support networkEndpointType: serviceEndpoint for SMB

### LINKED ISSUE #3303: Support networkEndpointType: serviceEndpoint for SMB (storage account firewall is currently NFS-only)

Today the driver can only restrict a dynamically provisioned storage account to a VNet when the volume is NFS. `updateSubnetServiceEndpoints()` is called from inside the NFS-only block in `CreateVolume`:

https://github.com/kubernetes-sigs/azurefile-csi-driver/blob/master/pkg/azurefile/controllerserver.go#L431

```go
var vnetResourceIDs []string
if fsType == nfs || protocol == nfs {
    ...
    if !ptr.Deref(createPrivateEndpoint, false) {
        // set VirtualNetworkResourceIDs for storage account firewall setting
        if vnetResourceIDs, err = d.updateSubnetServiceEndpoints(ctx, vnetResourceGroup, vnetName, subnetName); err != nil {
```

For SMB, `vnetResourceIDs` therefore stays empty and no `NetworkRuleSet` is ever set, so the account is created with `DefaultAction: Allow` (open to all networks).

That leaves only one way to get a network-restricted SMB account from dynamic provisioning: `networkEndpointType: privateEndpoint`. That is not usable in every environment. In ours, the central `privatelink.file.core.windows.net` zone is owned by a different team **in a different Azure subscription**, and the driver's private endpoint automation always builds the private DNS zone / vnet link / zone group IDs with the cluster's own subscription ID (`az.SubscriptionID` in `cloud-provider-azure`'s `azure_storageaccount.go`). There is no parameter to point at a zone in another subscription. When we granted the control plane identity DNS permissions, the driver simply created a **second, conflicting** `privatelink.file.core.windows.net` zone in our own resource group.

A VNet service endpoint + storage account firewall would solve this cleanly, has no DNS dependency at all, and Azure supports it perfectly well for SMB file shares. The driver just never wires it up for SMB.

Accept `serviceEndpoint` as a value for the existing `networkEndpointType` parameter, for any protocol:

```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: azurefile-csi-private
parameters:
  skuName: Standard_ZRS
  mountWithManagedIdentity: "true"
  networkEndpointType: "serviceEndpoint"
  vnetResourceGroup: "my-vnet-rg"
  vnetName: "my-vnet"
  subnetName: "my-aks-subnet"
```

This is additive only. Everything needed already exists and is protocol-agnostic:

| Component | Location | Protocol-specific? |
|---|---|---|
| `updateSubnetServiceEndpoints()` | `pkg/azurefile/azure.go` | No, only adds `Microsoft.Storage` to the subnet(s) |
| `NetworkRuleSet` / `VirtualNetworkRules` construction | `cloud-provider-azure` `azure_storageaccount.go` | No |
| `AreVNetRulesEqual()` account matching | `cloud-provider-azure` `azure_storageaccount.go` | No |
| `VNetResourceGroup` / `VNetName` / `SubnetName` in `AccountOptions` | `controllerserver.go` | No, already populated for SMB |
| **The call site** | **`controllerserver.go`** | **Yes, `if fsType == nfs \|\| protocol == nfs`** |

Note this must be **opt-in**, not on by default. #2277 shows a user who was unhappy about a service endpoint being created at all (it interfered with their UDR/next-hop-firewall routing). Gating it behind an explicit `networkEndpointType: serviceEndpoint` keeps current behaviour byte-for-byte identical for everyone who does not ask for it.

- `networkEndpointType: privateEndpoint` - blocked by the cross-subscription private DNS zone limitation described above.
- Pre-provisioning the storage account with Terraform and using static provisioning - works, and is what we are deploying now, but it gives up per-PVC dynamic provisioning and requires `SMBOauth` to be enabled out-of-band (the driver only sets `IsSmbOAuthEnabled` when it creates the account itself).
- Switching to NFS - not an option for Windows workloads, and Premium-only.

**Additional context**

Historically this looks accidental rather than deliberate. The call was introduced in commit `c95c785a` ("feat: update service endpoints in the subnets", Nov 2020) and was placed straight into the already existing NFS block, because NFS shares cannot be reached over the public endpoint at all and therefore *require* a secured network. SMB always had shared key / Kerberos auth as an alternative, so nobody needed it. I could not find any issue, comment or commit stating that this cannot or should not work for SMB.

I have a working patch with unit tests and will open a PR referencing this issue.

NielsG-it:
**What type of PR is this?**

/kind feature

Allows the storage account firewall to be configured for SMB volumes, by accepting `serviceEndpoint` as a value for the existing `networkEndpointType` parameter.

`updateSubnetServiceEndpoints()` was only called from inside the NFS-only branch of `CreateVolume`, so for SMB `vnetResourceIDs` always stayed empty and the storage account was created with `DefaultAction: Allow`. The only way to get a network restricted SMB account from dynamic provisioning was `networkEndpointType: privateEndpoint`, which is not usable when the `privatelink.file.core.windows.net` zone lives in a different subscription, because the private endpoint automation always resolves the private DNS zone in the cluster's own subscription (`az.SubscriptionID` in `cloud-provider-azure`).

Everything the feature needs was already protocol-agnostic (`updateSubnetServiceEndpoints()`, the `NetworkRuleSet` construction, `AreVNetRulesEqual()`, and the `VNetResourceGroup`/`VNetName`/`SubnetName` fields on `AccountOptions`, which are already populated for SMB). Only the call site was gated on NFS.

```yaml
parameters:
  skuName: Standard_ZRS
  mountWithManagedIdentity: "true"
  networkEndpointType: "serviceEndpoint"
  vnetResourceGroup: "my-vnet-rg"
  vnetName: "my-vnet"
  subnetName: "my-aks-subnet"
```

**Backwards compatibility**

Opt-in only, existing behaviour is unchanged:

- NFS without a private endpoint still always gets a service endpoint (moved, not changed).
- NFS with `privateEndpoint` still gets no service endpoint.
- SMB gets a service endpoint **only** when `networkEndpointType: serviceEndpoint` is set explicitly.

That last point is deliberate: in #2277 a user reported an *unwanted* service endpoint interfering with their UDR/next-hop-firewall routing, so this must never become default behaviour for SMB.

`createServiceEndpoint` is also added to the storage account `lockKey`, in the same spirit as #3111, so two storage classes that differ only in `networkEndpointType` cannot accidentally share an account or an `accountSearchCache` entry.

**Testing**

Three new specs in `controllerserver_test.go`:

1. SMB + `serviceEndpoint` updates the subnet and sets the vnet rules.
2. SMB without `networkEndpointType` never touches the subnet client.
3. SMB + `privateEndpoint` never touches the subnet client.

Full suite passes locally (`124 of 124 Specs, SUCCESS!`). I also verified the new test is meaningful by reverting the `controllerserver.go` change and re-running: spec 1 then fails on the unmet gomock expectation, while specs 2 and 3 still pass.

Fixes #3303

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Does this PR introduce a user-facing change?**

```release-note
feat: support `networkEndpointType: serviceEndpoint` in storage class to configure the storage account firewall for SMB volumes
```


andyzhangx:
hi @NielsG-it thanks for the contribution, can you sign the easy cla first?

andyzhangx (approved):
/lgtm

PR #3296, opened 2026-07-29
fix: ephemeral inline volume improvements

rnhan:
**What type of PR is this?**

/kind bug
/kind cleanup

**What this PR does / why we need it**:

Improvements to ephemeral inline volumes
- Improve service account secret access
  - Use podInfoOnMount and read the spec.serviceAccountName
  - Issue a SubjectAccessReview asking whether system:serviceaccount:<namespace>:<serviceaccount> can GET the referenced secret
  - Proceed with the mount only if allowed; otherwise fail with PermissionDenied.
  - off|warn|enforce options, default to off to not break existing setups
- Enforce unsupported features (e.g. NFS/VHD for inline ephemeral volumes)
- Validation for PV disk mounts
- Case collision validation for volumeContext/volumeAttributes
- Validate server/sharename mounting for inline ephemeral volumes

**Requirements**:
- [x] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [x] includes documentation
- [x] adds unit tests
- [x] tested upgrade from previous version

**Special notes for your reviewer**:

**Release note**:
```
none
```

PR #3308, opened 2026-07-31
Cherry pick of #3296 on release-1.35

rnhan:
Cherry pick of #3296 on release-1.35.

Includes the changes introduced by #3296, adapted to release-1.35:

- Improve service account secret access
  - Use podInfoOnMount and read the spec.serviceAccountName
  - Issue a SubjectAccessReview asking whether system:serviceaccount:<namespace>:<serviceaccount> can GET the referenced secret
  - Proceed with the mount only if allowed; otherwise fail with PermissionDenied.
  - off|warn|enforce options, default to off to not break existing setups
- Enforce unsupported features (e.g. NFS/VHD for inline ephemeral volumes)
- Validation for PV disk mounts
- Case collision validation for volumeContext/volumeAttributes
- Validate server/sharename mounting for inline ephemeral volumes

Adapted for release-1.35: the `serviceAccountToken` fast-path keeps release-1.35's `getValueInMap(context, serviceAccountTokenField)` form rather than master's `getServiceAccountTokens` / `hasStorageAccountCredentials` secrets refactor (not part of #3296).

/kind bug
/release-note-none

/cc @andyzhangx

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -37,6 +37,7 @@ require (
 	google.golang.org/protobuf v1.36.11
 	k8s.io/api v0.35.0
 	k8s.io/apimachinery v0.35.0
+	k8s.io/apiserver v0.35.0
 	k8s.io/client-go v0.35.0
 	k8s.io/cloud-provider v0.35.0
 	k8s.io/component-base v0.35.0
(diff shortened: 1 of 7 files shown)


andyzhangx (approved):
/lgtm

PR #3318, opened 2026-08-06
fix: time out Windows SMB unstage unmount

andyzhangx:
## Summary
This PR is being closed instead of merged.

## Why
After looking at the Windows unmount behavior again, the more likely issue is that application pods still have active access to the mounted Azure Files share when `NodeUnstageVolume` tries to remove the SMB global mapping.

In that situation, adding a driver-side timeout is not the right fix. A timeout would only make the unstage path fail faster, while the underlying problem would still be that the workload has not released the file share cleanly.

## Current recommendation
If Windows unmount gets stuck, the user should first resolve the application-side issue, for example by:
- identifying pods/processes that still access the mounted file share
- stopping or draining the affected workload/node so file handles are released
- retrying after the SMB mapping can be removed cleanly

Given that, this PR will be closed.

### DIFF
--- a/pkg/azurefile/fake_mounter.go
+++ b/pkg/azurefile/fake_mounter.go
@@ -19,6 +19,7 @@ package azurefile
 import (
 	"fmt"
 	"strings"
+	"time"
 
 	mount "k8s.io/mount-utils"
 )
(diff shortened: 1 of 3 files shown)


andyzhangx:
Closing this PR. The current conclusion is that a stuck Windows unmount is more likely due to application pods still accessing the mounted Azure Files share, and that should be resolved on the workload side rather than by adding a driver-side unmount timeout.

PR #3337, opened 2026-08-13
feat: support snapshot metadata in CreateSnapshot

mayankagg9722:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

Adds support for optional custom metadata in `VolumeSnapshotClass`. Users can provide comma-separated `key=value` pairs using the same format as the existing `tags` parameter, and the driver stores each entry on the Azure file share snapshot.

- New `metadata` parameter accepted in `CreateSnapshot` parameters, for example: `comment=snapshot before database migration,environment=production`.
- Reuses the existing `ConvertTagsToMap` helper to parse metadata into key-value pairs.
- Writes all custom metadata on both snapshot creation paths: data plane API (`share.CreateSnapshotOptions.Metadata`) and SRP (`armstorage.FileShareProperties.Metadata`).
- Reserves the driver-owned `initiator` metadata key, which is still added for snapshot idempotency.
- Preserves existing behavior when `metadata` is omitted: only the `initiator` key is set.
- Documents the parameter in `docs/driver-parameters.md`.
- Adds unit coverage for parsing, invalid metadata, the reserved key, and metadata propagation through both SRP and data-plane requests.

**Requirements**:

- [x] uses [conventional commit messages](https://www.conventionalcommits.org/)
- [x] includes documentation
- [x] adds unit tests

**Special notes for your reviewer**:

The parameter is optional and backward compatible. Metadata values are parsed with the existing tags format; Azure Files validates service-specific metadata name and size constraints.

**Release note**:

```release-note
feat: support custom metadata in VolumeSnapshotClass for Azure file share snapshots
```

PR #3343, opened 2026-08-17
test(e2e): add metadata parameter to snapshot storage class

andyzhangx:
**What type of PR is this?**

/kind test

Exercises the new `metadata` snapshot parameter (added in #3337) in the dynamic provisioning snapshot restore e2e test to make sure snapshot creation still works when custom metadata is supplied.

**Release note**:
```release-note
NONE
```


andyzhangx:
Addressed in db43ea36: added a new `VolumeSnapshotClassParameters` field on `DynamicallyProvisionedVolumeSnapshotTest` that merges into the `VolumeSnapshotClass` parameters before `CreateSnapshot`. Moved `metadata` there and reverted `SnapshotStorageClassParameters` back to volume-only parameters. Thanks @Copilot.

PR #3349, opened 2026-08-22
feat: add mount_operations_total metric to track per-share mount fail…

rnhan:
…ures

**What type of PR is this?**

/kind feature

Adds a metric to track mount failures to a single fileshare across multiple nodes and clusters, including for retries, with bucketed reasons.

- Dedicated `azurefile_csi_driver_mount_operations_total` counter, emitted at 3 mount command locations in `NodeStageVolume`, with the target share's `storage_account`, `file_share`, `subscription_id`, and `resource_group` along with `protocol`, `success`, and classified `mount_error_reason`
  - Allows a single fileshare with repeated errors to be tracked across multiple nodes/clusters in a subscription
  - Reason is bucketed by `classifyMountError` into `timeout`, `enoent`, `access_denied`, `network`, `stale`, `busy`, `other` from the mount command's `mount.cifs` or `mount.nfs` stderr
  - `storage_account` and `file_share` are only recorded on failures (left blank on successes) with all identity values lowercased, in order to bound cardinality to the set of distinct failing shares.  

**Requirements**:
  
- [x] includes documentation
- [x] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/deploy/example/metrics/README.md
+++ b/deploy/example/metrics/README.md
@@ -49,6 +49,7 @@ The Azure File CSI driver exposes the following custom metrics:
 | `azurefile_csi_driver_operation_duration_seconds` | Histogram | `operation`, `success` | Duration of CSI operations in seconds |
 | `azurefile_csi_driver_operations_total` | Counter | `operation`, `success` | Total number of CSI operations |
 | `azurefile_csi_driver_mount_operations_total` | Counter | `operation`, `success`, `protocol`, `storage_account`, `file_share`, `subscription_id`, `resource_group`, `mount_error_reason` | Total number of mount-path executions, labeled with the target share identity and a classified error reason. Used to track mount failures for a single file share across many nodes/clusters/subscriptions. |
+| `azurefile_csi_driver_inline_volume_operations_total` | Counter | `success` | Total number of ephemeral (inline) volume `NodePublishVolume` operations. Incremented only on the inline code path. |
 
 **Label Values:**
 - `operation`: `node_stage_volume`, `node_unstage_volume`, `node_publish_volume`, `node_unpublish_volume`
--- a/deploy/example/metrics/README.md
+++ b/deploy/example/metrics/README.md
@@ -48,19 +48,19 @@ The Azure File CSI driver exposes the following custom metrics:
 |--------|------|--------|-------------|
 | `azurefile_csi_driver_operation_duration_seconds` | Histogram | `operation`, `success` | Duration of CSI operations in seconds |
 | `azurefile_csi_driver_operations_total` | Counter | `operation`, `success` | Total number of CSI operations |
-| `azurefile_csi_driver_mount_operations_total` | Counter | `operation`, `success`, `protocol`, `storage_account`, `file_share`, `subscription_id`, `resource_group`, `mount_error_reason` | Total number of mount-path executions, labeled with the target share identity and a classified error reason. Used to track mount failures for a single file share across many nodes/clusters/subscriptions. |
-| `azurefile_csi_driver_inline_volume_operations_total` | Counter | `success` | Total number of ephemeral (inline) volume `NodePublishVolume` operations. Incremented only on the inline code path. |
+| `azurefile_csi_driver_mount_operations_total` | Counter | `operation`, `success`, `protocol`, `storage_account`, `subscription_id`, `mount_error_reason` | Total number of mount-path executions, labeled with the target storage account identity (on failures) and a classified error reason. Used to track mount failures for a single file share across many nodes/clusters/subscriptions. |
 
 **Label Values:**
-- `operation`: `node_stage_volume`, `node_unstage_volume`, `node_publish_volume`, `node_unpublish_volume`
+- `operation`: `node_stage_volume`, `node_unstage_volume`, `node_publish_volume`, `node_unpublish_volume`, `inline_mount`
 - `success`: `true`, `false`
+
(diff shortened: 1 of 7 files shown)


pkg/azurefile/nodeserver.go
```diff
@@ -615,12 +621,16 @@ func (d *Driver) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRe
 		}
 		if mountFsType == aznfs {
 			klog.V(2).Infof("either of encryptInTransit (%t) (or) useAZNFSForNFSMounts (%t) is enabled, mount by azurefile-proxy", encryptInTransit, d.useAZNFSForNFSMounts)
-			if err := d.mountWithProxy(ctx, source, cifsMountPath, mountFsType, mountOptions, sensitiveMountOptions); err != nil {
-				if strings.Contains(err.Error(), "no such file or directory") {
+			mountMC := csiMetrics.NewCSIMetricContext("node_stage_volume_mount").WithBasicVolumeInfo(rgName, subsID, d.Name)
+			mountErr := d.mountWithProxy(ctx, source, cifsMountPath, mountFsType, mountOptions, sensitiveMountOptions)
+			mountMC.ObserveMountWithLabels(mountErr == nil, csiMetrics.Protocol, mountFsType, csiMetrics.StorageAccount, accountName, csiMetrics.FileShare, shareBaseName,
+				csiMetrics.SubscriptionID, subsID, csiMetrics.ResourceGroup, rgName, csiMetrics.MountErrorReason, classifyMountError(mountErr))
```
mittachaitu: Is it good idea to push original error as part of metrics? Usually it is bad idea to push error msg as part of metrics since we are also pushing lot other information so checking on original error msg.

pkg/azurefile/nodeserver.go
```diff
@@ -615,12 +621,16 @@ func (d *Driver) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRe
 		}
 		if mountFsType == aznfs {
 			klog.V(2).Infof("either of encryptInTransit (%t) (or) useAZNFSForNFSMounts (%t) is enabled, mount by azurefile-proxy", encryptInTransit, d.useAZNFSForNFSMounts)
-			if err := d.mountWithProxy(ctx, source, cifsMountPath, mountFsType, mountOptions, sensitiveMountOptions); err != nil {
-				if strings.Contains(err.Error(), "no such file or directory") {
+			mountMC := csiMetrics.NewCSIMetricContext("node_stage_volume_mount").WithBasicVolumeInfo(rgName, subsID, d.Name)
+			mountErr := d.mountWithProxy(ctx, source, cifsMountPath, mountFsType, mountOptions, sensitiveMountOptions)
+			mountMC.ObserveMountWithLabels(mountErr == nil, csiMetrics.Protocol, mountFsType, csiMetrics.StorageAccount, accountName, csiMetrics.FileShare, shareBaseName,
```
mittachaitu: Will we able to get timeout error if call to/at proxy hangs?

mittachaitu (commented):
Provided small comments

pkg/azurefile/nodeserver.go
rnhan: Yes, updated to be certain

pkg/azurefile/nodeserver.go
rnhan: I think we can do this here, since it's an error type that should help with classifying for dashboards and alerts to minimize noise, because most of the other information to triage is pushed separately as node logs.  Let me know what you think.

mittachaitu: Make sense!!

mittachaitu (approved):
Thanks for having metrics for inline volume, LGTM

deploy/example/metrics/README.md
```diff
@@ -48,11 +48,22 @@ The Azure File CSI driver exposes the following custom metrics:
 |--------|------|--------|-------------|
 | `azurefile_csi_driver_operation_duration_seconds` | Histogram | `operation`, `success` | Duration of CSI operations in seconds |
 | `azurefile_csi_driver_operations_total` | Counter | `operation`, `success` | Total number of CSI operations |
+| `azurefile_csi_driver_mount_operations_total` | Counter | `operation`, `success`, `protocol`, `storage_account`, `subscription_id`, `mount_error_reason` | Total number of mount-path executions, labeled with the target storage account identity (on failures) and a classified error reason. Used to track mount failures for a single file share across many nodes/clusters/subscriptions. |
```
landreasyan, line 51: since we are tracking mount operations in this metrics, is the operations redundant? Otherwise, wondering if we can expend and use the `azurefile_csi_driver_operations_total` for tracking mount operations as well?

deploy/example/metrics/README.md
```diff
@@ -48,11 +48,22 @@ The Azure File CSI driver exposes the following custom metrics:
...
+| `azurefile_csi_driver_inline_volume_operations_total` | Counter | `success` | Total number of ephemeral (inline) volume `NodePublishVolume` operations. Incremented only on the inline code path. |
```
landreasyan: Do we need a new counter for this or can we track it through `azurefile_csi_driver_operations_total` through, say `inline_mount` operation?

deploy/example/metrics/README.md
rnhan: Doing this would require us add ingthose additional storage account and subscription id labels to ALL other metrics, which I think we should avoid doing

deploy/example/metrics/README.md
rnhan: Good idea, done

PR #3360, opened 2026-08-24
docs: clarify inline (ephemeral) CSI volumes are not supported for managed identity mount

andyzhangx:
**What type of PR is this?**
/kind documentation

Clarify in `docs/managed-identity-mount.md` that **inline (ephemeral) CSI volumes are not a supported path** for managed identity authentication.

The node driver already rejects `mountWithManagedIdentity: "true"` on ephemeral volumes ([`pkg/azurefile/nodeserver.go`](https://github.com/kubernetes-sigs/azurefile-csi-driver/blob/master/pkg/azurefile/nodeserver.go#L118-L120)) so tenant Pods cannot mount arbitrary file shares under the node's identity, but the docs do not currently call this out. Users occasionally try inline volume + managed identity and hit an `InvalidArgument` at NodePublish time without a clear pointer to the supported StorageClass/PV path.

> Note: this PR intentionally does **not** touch `docs/workload-identity-static-pv-mount.md` — inline volumes with `mountWithWorkloadIdentityToken: "true"` **are** supported by the node driver (see the `useWIToken` path following the MI/OAuth guard).

Docs-only clarification, no behavior change.

**Does this PR introduce a user-facing change?**:
```release-note
NONE
```

### DIFF
--- a/docs/managed-identity-mount.md
+++ b/docs/managed-identity-mount.md
@@ -6,7 +6,7 @@
 This article demonstrates how to mount an SMB file share using user-assigned managed identity authentication, without relying on account key authentication.
 
 > [!IMPORTANT]
-> **Inline (ephemeral) CSI volumes are not supported.** Requests with `mountWithManagedIdentity: "true"` on a `csi:` inline volume in a Pod spec are rejected by the node driver. Use a StorageClass (dynamic) or PersistentVolume (static) so the identity and `clientID` are controlled by the cluster admin, not by tenant Pods. This prevents arbitrary Pods from mounting file shares under the node's managed identity.
+> **Managed identity mount is not supported for inline (ephemeral) CSI volumes.** Requests with `mountWithManagedIdentity: "true"` on a `csi:` inline volume in a Pod spec are rejected by the node driver; use a StorageClass (dynamic) or PersistentVolume (static) so the identity and `clientID` are controlled by the cluster admin, not by tenant Pods. Inline volumes remain supported with workload identity (`mountWithWorkloadIdentityToken: "true"`) or secret-based authentication.
 
 > [!NOTE]
 > By default, you can leverage the built-in user-assigned managed identity (kubelet identity) bound to the AKS agent node pool (with the naming convention [`<AKS Cluster Name>-agentpool`](https://docs.microsoft.com/en-us/azure/aks/use-managed-identity#summary-of-managed-identities)).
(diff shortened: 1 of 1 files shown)


docs/workload-identity-static-pv-mount.md
```diff
@@ -5,6 +5,7 @@
 ## Limitations
 
 - This feature is **not supported for NFS mounts** since NFS does not require credentials.
+- **Inline (ephemeral) CSI volumes are not supported.** Requests with `mountWithWorkloadIdentityToken: "true"` on a `csi:` inline volume in a Pod spec are rejected. Use a StorageClass (dynamic) or PersistentVolume (static) as shown below so the identity and `clientID` are controlled by the cluster admin.
```
mittachaitu: We are blocking support of `mountWithOAuthTokenField` not `mountWithWorkloadIdentityToken`

andyzhangx: it's already blocked (not related to your PR):
```
			// When Managed Identity or OAuth token is used for ephemeral volumes then reject the request.
			// Allowing access for inline volume with identity will open up risk of arbitrary pods accessing fileshares with node identity permissions.
			if strings.EqualFold(getValueInMap(context, mountWithManagedIdentityField), trueValue) {
				return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("mountWithManagedIdentity cannot be used for ephemeral volumes, please use either %s or secret based authentication", mountWithWITokenField))
			}
			if strings.EqualFold(getValueInMap(context, mountWithOAuthTokenField), trueValue) {
				return nil, status.Error(codes.InvalidArgument, "mountWithOAuthToken cannot be used for ephemeral volumes, please use secret based authentication")
			}
```

andyzhangx: fixed

docs/managed-identity-mount.md
```diff
@@ -5,6 +5,9 @@
 
 This article demonstrates how to mount an SMB file share using user-assigned managed identity authentication, without relying on account key authentication.
 
+> [!IMPORTANT]
+> **Inline (ephemeral) CSI volumes are not supported.** Requests with `mountWithManagedIdentity: "true"` on a `csi:` inline volume in a Pod spec are rejected by the node driver. Use a StorageClass (dynamic) or PersistentVolume (static) so the identity and `clientID` are controlled by the cluster admin, not by tenant Pods. This prevents arbitrary Pods from mounting file shares under the node's managed identity.
```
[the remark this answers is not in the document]
andyzhangx: Good catch — reworded the callout to scope it to managed identity mount only and explicitly note that inline volumes are still supported with workload identity (`mountWithWorkloadIdentityToken: "true"`) or secret-based authentication. Fixed in e155638.

PR #3365, opened 2026-08-27
docs: clarify that allowSharedKeyAccess applies to both SMB and NFS

andyzhangx:
## What type of PR is this?
/kind documentation

## What this PR does / why we need it:

The `allowSharedKeyAccess` parameter was listed under **"Following parameters are only for NFS protocol"**, but the driver actually forwards it to the storage account creation call regardless of protocol (see `pkg/azurefile/controllerserver.go` where `AllowSharedKeyAccess` is unconditionally passed through into `AccountOptions`). Users on SMB workloads read the current wording as *"this option is silently ignored on SMB"*, which is misleading.

SMB workloads can also disable shared key access — but only when:

1. `storeAccountKey=false` is set (the controller explicitly rejects the default `storeAccountKey=true` when shared key access is disabled — see `controllerserver.go` around the `storeAccountKey is not supported for account with shared access key disabled` error), **and**
2. Identity-based SMB mount is configured (see [managed-identity-mount.md](https://github.com/kubernetes-sigs/azurefile-csi-driver/blob/master/docs/managed-identity-mount.md)).

Without both, SMB provisioning/mount will fail once the account has shared key access disabled.

This PR:
- Moves `allowSharedKeyAccess` out of the "NFS-only" section into a new "applies to both SMB and NFS protocols" section.
- Adds explicit notes for both protocols so users understand what's required to disable shared key access on SMB.

## Which issue(s) this PR fixes:
None (documentation clarification).

## Requirements:
- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests
  - N/A (docs only)

## Release note:
```release-note
NONE
```


andyzhangx:
Updated the SMB note in 60e1a73 to reflect the actual controller behavior.

Correction: users do **not** need to manually set `storeAccountKey=false`. The controller (`controllerserver.go` `if mountWithManagedIdentity || mountWithWIToken || mountWithOAuthToken` block) already sets `storeAccountKey=false` and `requiresSmbOAuth=true` automatically whenever any of `mountWithManagedIdentity`, `mountWithWorkloadIdentityToken`, or `mountWithOAuthToken` is true. So on SMB the user-facing contract is: opt into one of those identity-based mount flags, and the driver handles the rest.

PR #3367, opened 2026-08-28
feat: add support for data plane metrics for linux environment

mittachaitu:
**What type of PR is this?**
/kind feature

This PR adds support for data plane metrics by exporting information
present under `/proc/fs/cifs/Stats` and `/proc/self/mountstats` for both
cifs and nfs mounts.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**How it works**:
- Data plane metrics are exported only when some external component hit `/metrics` request of csi-node driver.
- It works on pull based mode and doesn't scrape itself.
- Sample output:
```sh
azurefile_csi_driver_volume_read_bytes_total{file_share="pvc-0b8ea6a8-1ea9-45cf-bb30-3383ba115c8d",protocol="smb",storage_account="f389ac36a70ed4cb5bc601e"} 0
azurefile_csi_driver_volume_read_bytes_total{file_share="pvc-bfd5f1c9-cd69-4bea-b77c-953d9dcded9b",protocol="smb",storage_account="f389ac36a70ed4cb5bc601e"} 1.4143488e+07
azurefile_csi_driver_volume_read_bytes_total{file_share="pvc-efe87cbb-2a3d-4af0-81ce-8e51986153a7",protocol="smb",storage_account="f389ac36a70ed4cb5bc601e"} 7.5808768e+07
azurefile_csi_driver_volume_read_bytes_total{file_share="pvcn-4da283f5-254b-4e25-91c0-f1e022fdd536",protocol="nfs",storage_account="f33b7f3e834cd464fb38cd7"} 3.42474752e+08
# HELP azurefile_csi_driver_volume_stats_collector_up [ALPHA] Whether the collector successfully read a statistics source.
# TYPE azurefile_csi_driver_volume_stats_collector_up gauge
azurefile_csi_driver_volume_stats_collector_up{source="nfs"} 1
azurefile_csi_driver_volume_stats_collector_up{source="persistentvolumes"} 1
azurefile_csi_driver_volume_stats_collector_up{source="smb"} 1
# HELP azurefile_csi_driver_volume_written_bytes_total [ALPHA] Total bytes written to the remote filesystem.
# TYPE azurefile_csi_driver_volume_written_bytes_total counter
azurefile_csi_driver_volume_written_bytes_total{file_share="pvc-0b8ea6a8-1ea9-45cf-bb30-3383ba115c8d",protocol="smb",storage_account="f389ac36a70ed4cb5bc601e"} 0
azurefile_csi_driver_volume_written_bytes_total{file_share="pvc-bfd5f1c9-cd69-4bea-b77c-953d9dcded9b",protocol="smb",storage_account="f389ac36a70ed4cb5bc601e"} 1.079672832e+09
azurefile_csi_driver_volume_written_bytes_total{file_share="pvc-efe87cbb-2a3d-4af0-81ce-8e51986153a7",protocol="smb",storage_account="f389ac36a70ed4cb5bc601e"} 3.2432128e+07
azurefile_csi_driver_volume_written_bytes_total{file_share="pvcn-4da283f5-254b-4e25-91c0-f1e022fdd536",protocol="nfs",storage_account="f33b7f3e834cd464fb38cd7"} 2.285299838e+09
```

**Release note**:
```
none
```


pkg/azurefile/stats/volume/remote_volume.go
```diff
@@ -0,0 +1,113 @@
...
+		client: client,
+	}
+}
+
+func (r *RemoteVolume) List(ctx context.Context) (MetadataList, error) {
+	if r == nil || r.client == nil {
+		return nil, fmt.Errorf("kubernetes client is nil")
+	}
+
+	persistentVolumes, err := r.client.CoreV1().PersistentVolumes().List(ctx, metav1.ListOptions{})
```
andyzhangx: this is very expensive call when there are thousands of PVs, may lead to OOM

andyzhangx: and why do you need PV list on the node, that would also require additional permission.

mittachaitu: > this is very expensive call when there are thousands of PVs, may lead to OOM. 

Agree!!

> and why do you need PV list on the node, that would also require additional permission.

I thought of collecting metrics only for volumes belongs to `files.csi.driver` instead of projecting all `cifs/nfs` mounts on node as from csi driver.

mittachaitu: I will remove dependencies on Kubernetes persistent volume

kapilupadhayay:
1. Go Lint is failing (already red on the PR):
•  pkg/azurefile/stats/stats_linux_test.go:93  — misspell:  other-share  triggers  ther → there . Rename the fixture (e.g.  second-share ).
•  pkg/azurefile/stats/stats_linux.go:67  — revive stutter:  stats.StatsCollection  → rename to  Collection .
2. Helm chart tgz not regenerated.  charts/latest/azurefile-csi-driver/templates/rbac-csi-azurefile-node.yaml  changed but  charts/latest/azurefile-csi-driver-v0.0.0.tgz  still lacks the  persistentvolumes  rule (verified by extracting it).  hack/verify-helm-chart-files.sh  will fail. Run  helm package charts/latest/azurefile-csi-driver -d charts/latest/ .
A few small things:
3.  pkg/azurefile/stats/smb_linux_stats.go:247  — local variable named  Device  (also at line ~168); lowercase it to match Go convention.
4.  pkg/azurefile/stats/nfs_stats.go  has no  _linux  suffix or build tag, while its test is  nfs_stats_linux_test.go  and everything else in the package is  _linux -suffixed. Renaming to  nfs_stats_linux.go  would be more consistent.
5. Cardinality: 14 SMB operations × {requests, errors} × shares-per-node is ~28 extra series per mounted share on top of the 2 byte counters. Consider gating the per-operation metrics behind a flag (default off) so users who only want throughput don't pay for it.
6. The PR checklist has "includes documentation" unchecked — this adds a user-visible metrics surface, so a short doc page listing metric names, labels, stability level, and the Linux-only caveat would help.

pkg/azurefile/stats/stats_linux.go
```diff
@@ -0,0 +1,223 @@
...
+
+	volumeTargets := newVolumeTargetSet(volumes)
+	for _, stat := range cifsStats {
+		account, share, ok := parseCIFSTarget(stat.Device)
+		if ok && volumeTargets.contains(account, share) {
+			collection.add(projectCIFSStats(stat, account, share), filesystemIndexes)
+		}
+	}
+	for _, stat := range nfsStats {
+		account, share, ok := parseNFSTarget(stat.Device)
```
kapilupadhayay: CIFS path has no mount-path filter (NFS does) — matters more once the PV lookup is removed

pkg/azurefile/stats/smb_linux_stats.go
```diff
@@ -0,0 +1,289 @@
...
+		// Global and transport-level fields occur outside and between share
+		// records. They cannot be attributed to an individual share.
+		if current == nil {
+			continue
+		}
+
+		switch {
+		case strings.HasPrefix(line, "Bytes read:"):
+			if _, err := fmt.Sscanf(line, "Bytes read: %d Bytes written: %d", &current.BytesRead, &current.BytesWritten); err != nil {
+				return nil, parseError(cifsStatsPath, lineNumber, line, err)
```
kapilupadhayay: One malformed line aborts all SMB stats → collector_up{smb}=0; skip-and-continue instead

pkg/azurefile/stats/smb_linux_stats.go
```diff
@@ -0,0 +1,289 @@
...
+	// OplockBreaksFailed is the number of failed SMB OPLOCK_BREAK messages.
+	OplockBreaksFailed uint64
+}
+
+func ProcessCIFSStats(cifsStatsPath string) ([]CIFSStats, error) {
+	if strings.TrimSpace(cifsStatsPath) == "" {
+		return nil, fmt.Errorf("cifsStatsPath is empty")
+	}
+
+	file, err := os.Open(cifsStatsPath)
```
kapilupadhayay, line 142: Missing /proc/fs/cifs/Stats on NFS-only nodes = per-scrape error spam; treat os.ErrNotExist as empty

pkg/azurefile/stats/collector_linux.go
```diff
@@ -0,0 +1,145 @@
...
+	ch <- c.requests
+	ch <- c.errors
+	ch <- c.up
+}
+
+func (c *VolumeStatsCollector) CollectWithStability(ch chan<- metrics.Metric) {
+	ctx, cancel := context.WithTimeout(context.Background(), collectionTimeout)
+	defer cancel()
+
+	volumes, err := c.volumes.List(ctx)
```
rnhan: Scrape timeout is 10s here, but the AKS scrape config timeout I believe is also 10 seconds, so if PV list times out or takes 9 seconds then if the /proc reads take longer than 1 second the total CollectWithStability is longer than the scrape timeout, and the scrape fails.  I think you mentioned removing dependency on the list PV operation though since it's a very expensive operation, so might be less applicable, but would be good to make sure we avoid exceeding the prometheus configs and failing the scrapes entirely

pkg/azurefile/stats/stats_linux.go
```diff
@@ -0,0 +1,223 @@
...
+		return "", "", false
+	}
+
+	parts := strings.Split(strings.Trim(export, "/"), "/")
+	if len(parts) < 2 {
+		return "", "", false
+	}
+
+	// Azure Files NFS exports use /<storage-account>/<share>.
+	return normalizeTarget(server, parts[len(parts)-1])
```
rnhan: Do we need to be concerned here about /storageaccount/fileshare/subdir down here?

pkg/azurefile/stats/stats_linux.go
```diff
@@ -0,0 +1,223 @@
...
+
+func isAzureFileCSIGlobalMount(mountPoint string) bool {
+	const driverMountPath = "/plugins/kubernetes.io/csi/file.csi.azure.com/"
+
+	clean := filepath.ToSlash(filepath.Clean(mountPoint))
+	return strings.Contains(clean, driverMountPath) &&
+		strings.HasSuffix(clean, "/globalmount")
+}
+
+func projectCIFSStats(stat CIFSStats, account, share string) FilesystemStats {
```
rnhan: 14 operations * 2 requests/total * # fileshares attached to the node, is 28 metric series per SMB share (on the node) going to cause memory issues in clusters with lots of fileshares?   Wondering how aggressively this grows especially over time, but also not sure of a good solution

pkg/azurefile/stats/stats_linux.go
mittachaitu: I have updated the code to look at `/var/lib/kubelet/plugin/azure-file-csi` for global mounts and for pod mounts, (possible to have only pod mounts when Workload Identity is enabled) we will look at `vol_data.json`

pkg/azurefile/stats/volume/remote_volume.go
mittachaitu: Updated!!

pkg/azurefile/stats/stats_linux.go
mittachaitu: Updated the code!!

pkg/azurefile/stats/smb_linux_stats.go
mittachaitu: Updated the code

pkg/azurefile/stats/smb_linux_stats.go
```diff
@@ -0,0 +1,295 @@
...
+	if err != nil {
+		if os.IsNotExist(err) {
+			klog.Warningf("cifs stats file %s does not exist", cifsStatsPath)
+			return nil, nil
+		}
```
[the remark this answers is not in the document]
mittachaitu: Showing empty should be ok and in system having only nfs module will make sense to treat as non-error

PR #3384, opened 2026-09-02
fix: invalidate account key cache on SMB mount failure

andyzhangx:
**What type of PR is this?**
/kind bug

When a customer rotates their Azure storage account key while a pod is running, the driver's account key cache (`accountCacheMap`, `TimedCache` with a fixed ~3min TTL — reads do NOT extend the TTL) can keep handing back the stale key for up to the full TTL window. During that window, every kubelet `NodeStageVolume` retry uses the stale key and `mount.cifs` fails with `Permission denied`; the mount only recovers once the entry naturally ages out.

Repro reported from field:
1. Kubelet retries `NodeStageVolume` with exponential backoff (cap ~2min).
2. The retries fall inside the ~3min cache TTL window.
3. Every retry reads the stale key from `accountCacheMap` and calls `mount.cifs`, which fails.
4. Recovery only happens when the entry ages out (or the pod is deleted and the driver takes a different code path).

**Fix**: on any SMB mount failure inside `NodeStageVolume`, invalidate the `accountCacheMap` entry for that account when the mount actually used the account key. The next retry goes through `GetStorageAccesskey` → `GetStorageAccesskeyWithSubsID` and picks up the rotated key from ARM. This shortens the recovery window from "up to TTL" to "one retry".

**Scope**:
- Covers Linux SMB (`mount.cifs`) and Windows SMB (host-process container deployment, the default today — `SMBMount` authenticates with the same cached account key), so both suffer the same stale-cache window after rotation.
- Skipped for auth modes that do not read the account key from the cache:
    * NFS (Kerberos / private endpoint / mount options).
    * SMB with managed identity, workload-identity token, or OAuth token (Kerberos).
- On mount success — no change (cache is untouched).
- On mount failure with a still-valid key (e.g. transient network / SMB issue) — we drop and re-fetch the same key on the next retry, which is cheap and self-healing.

Fixes #

Related docs (customer workaround so far): https://learn.microsoft.com/en-us/troubleshoot/azure/azure-kubernetes/storage/file-share-mount-failures-azure-files

**Release note**:
```release-note
Fix stale account key cache after storage account key rotation — the driver now invalidates the cached account key when an SMB mount attempt fails, so subsequent retries pick up a rotated key without waiting for the cache TTL to expire.
```

### DIFF
--- a/pkg/azurefile/nodeserver.go
+++ b/pkg/azurefile/nodeserver.go
@@ -685,6 +685,23 @@ func (d *Driver) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRe
 					csiMetrics.SubscriptionID, subsID, csiMetrics.MountErrorReason, classifyMountError(mountErr))
 			}
 			if mountErr != nil {
+				// Invalidate the cached account key so the next mount retry
+				// re-fetches a fresh key. When a customer rotates their
+				// storage account key, the driver's account key cache
+				// (default TTL 3min, and refreshed on every read) can keep
+				// handing back the stale key for the lifetime of the pod,
+				// because every kubelet retry re-reads and extends the
+				// entry. Dropping it here breaks that loop: the next
+				// NodeStageVolume goes through GetStorageAccesskey ->
+				// GetStorageAccesskeyWithSubsID and picks up the new key.
+				if accountName != "" && d.accountCacheMap != nil {
+					if dErr := d.accountCacheMap.Delete(accountName); dErr != nil {
+						klog.Warningf("NodeStageVolume: failed to invalidate account key cache for %s after mount failure: %v", accountName, dErr)
+					} else {
+						klog.V(2).Infof("NodeStageVolume: invalidated account key cache for %s after mount failure on volume %s", accountName, volumeID)
+					}
+				}
+
 				var helpLinkMsg string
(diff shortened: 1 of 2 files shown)


andyzhangx:
Addressed Copilot review in `0ec098f`:

1. **Scope guard broadened** — invalidation now skipped when the mount is not using the account key at all:
   - `protocol == nfs` (Kerberos / private endpoint / mount options)
   - `mountWithManagedIdentity` (Kerberos via MI)
   - `mountWithWIToken` (Kerberos via workload identity)
   - non-empty `clientID`

   Only plain account-key SMB mounts (Linux CIFS + Windows `New-SmbGlobalMapping`, both of which take `username=<account>,password=<key>`) drop the cache on failure — which is exactly the population the rotate-key mitigation targets.

2. **Regression test** `TestInvalidateAccountKeyCacheOnMountFailure` in `pkg/azurefile/nodeserver_test.go` covers:
   - pre-condition: cache seeded
   - success path: entry preserved
   - failure path: entry dropped
   - post-rotation: re-seed surfaces the new key, not the stale one

   so the safety guarantee (successful mount never evicts) is asserted.

andyzhangx:
addressed by https://github.com/kubernetes-sigs/azurefile-csi-driver/pull/3406

PR #3429, opened 2026-09-17
fix: re-relax inline volume restrictions for future deprecation

rnhan:
**What type of PR is this?**
/kind bug

Re-relaxes ephemeral (inline) volume validation that was rejecting valid configurations. Allows the NFS protocol for inline volumes again, and removes the mount-source check that rejected valid `server`/`shareName` values (e.g. NFS `/account/share` and SMB `share/subdir`). These restrictions are relaxed ahead of the planned deprecation of the inline volume path.

Fixes #

**Requirements**:
- [ ] includes documentation
- [x] adds unit tests

**Release note**:
```
Re-relaxed inline (ephemeral) volume restrictions: NFS protocol and path-separated server/shareName values are accepted again.
```


andyzhangx (approved):
/lgtm

PR #3430, opened 2026-09-17
fix: re-allow VHD for inline volumes for future deprecation

rnhan:
**What type of PR is this?**
/kind bug

Re-allows the VHD disk feature (`diskName` / disk `fsType`) for ephemeral (inline) volumes, which was rejected by a recent validation change. This restriction is relaxed ahead of the planned deprecation of the inline volume path.

Fixes #

**Requirements**:
- [ ] includes documentation
- [x] adds unit tests

**Release note**:
```
Re-allowed VHD (diskName / disk fsType) for inline (ephemeral) volumes.
```


andyzhangx (approved):
/lgtm

